PR #1, opened 2019-09-30
Import template tests from pennylane

johannesjmeyer:
Import the tests from the pennylane plugin template and make them work for cirq.

Issue: #2


johannesjmeyer:
Hi @antalszava -- thanks for the review! I suppose the tests were failing because my local pennylane version is the one directly from git master that has some new functionality compared to 0.5.

I'll apply your changes and will then merge!

Issue #6, opened 2019-10-17
Revise sampling behaviour in analytic mode


johannesjmeyer:
Currently, analytic mode samples with numpy not using the underlying Cirq measurement approach

johannesjmeyer:
Additionally, support multi-wire observables like `Hermitian`

PR #11, opened 2019-10-20
Multi wire observables

johannesjmeyer:
Adds support for multi-wire observables (qml.Hermitian), also laying the groundwork for the implementation of tensor observables.

Issues: #6


pennylane_cirq/cirq_device.py
```diff
@@ -60,20 +61,22 @@ class CirqDevice(Device):
...
             (np.array[float]): the converted measurements
         """
-        conversion = np.vectorize(lambda x: one_value if x else zero_value)
+        N = measurements.shape[0]     
+
+        indices = np.ravel_multi_index(measurements, [2] * N)
+        converted_measurements = eigenvalues[indices]
 
-        return conversion(measurements.flatten())
+        return converted_measurements
```
josh146: Nice!

pennylane_cirq/simulator_device.py
```diff
@@ -42,6 +43,15 @@
 from .cirq_device import CirqDevice
 
 
+@functools.lru_cache()
+def z_eigs(n):
+    """Return the eigenvalues of an n-fold tensor product of Pauli Z operators."""
```
josh146: This function docstring is better than the one in the qiskit plugin, might copy it back :laughing:

pennylane_cirq/simulator_device.py
```diff
@@ -98,11 +108,15 @@ def apply(self, operation, wires, par):
 
             if len(basis_state_array) != len(self.qubits):
```
josh146: Just a comment out of curiosity, but you decided to go with the `apply`, `measure()` etc. route, rather than overwrite `execute()`? I find nowadays that `apply`, `post_apply`, `measure`, etc. tend to be bit restrictive, leading to odd plugin design.

pennylane_cirq/simulator_device.py
```diff
@@ -198,95 +216,84 @@ def marginal_probability(self, wires):
...
-        if observable == "Hermitian":
-            # Take the eigenvalues from the stored values
-            Hmat = par[0]
-            Hkey = tuple(Hmat.flatten().tolist())
-            zero_value = self._eigs[Hkey]["eigval"][0]
-            one_value = self._eigs[Hkey]["eigval"][1]
-
-        elif observable == "Identity":
-            one_value = 1
+        eigenvalues = self._get_eigenvalues(observable, wires, par)
```
josh146: This is great @johannesjmeyer! I think it's cleaner than the Qiskit structure. I'm thinking that since so many plugins are landing on using the same methods/functions, we should move them to PennyLane as a convenience.

We could even have a `QubitDevice` base class in PennyLane, that implements the 'agnostic' logic. Things like the `rotate_circuit`, which calls `dev.apply()` to rotate wires pre-measurement.

tests/test_simulator_device.py
```diff
@@ -424,3 +424,108 @@ def test_expval_two_wires_with_parameters(self, simulator_device_2_wires, tol, o
...
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.var(qml.PauliX(0))
+
+        var = circuit()
+
+        # With 3 samples we are guaranteed to see a difference between
+        # an estimated variance an an analytically calculated one
+        assert var != 1.0
```
josh146:
```suggestion
        assert var != 1.0

```

pennylane_cirq/simulator_device.py
johannesjmeyer: Actually I just oriented myself at the other plugins and those mostly follow the `apply`, `measure`, etc. pattern. I think the code will become more lightweight if we move all those `BasisState` and `QubitStateVector` checks to PL core and I can remove them here.

But you're right, in the next plugin I will probably just overwrite `execute`

pennylane_cirq/simulator_device.py
johannesjmeyer: Yes, I think too that there is some unified structure emerging, especially in the analytic plugins. The process of going from the state vector/probabilities to expectation values, variances and samples is essentially the same. 

But I would not (yet) create a new kind of `Device` for this. If we offer some convenience functions like `utils.expval(probabilities, eigenvalues)` then we can have a unified structure where it is appropriate. 

For the rotations, I would probably add them to the `Observables` themselves as `diagonalize()` (creating PL operations like the MottonenStatePrep) or something alike, so that plugins can call them if they only support Z expectations. But I would not create an automatism because that may not be appropriate for some plugins.

PR #12, opened 2019-10-23
Increase Coverage

johannesjmeyer:
This Pull Request prepares the initial release of PL-Cirq.

It increases the coverage.

### DIFF
--- a/pennylane_cirq/ops.py
+++ b/pennylane_cirq/ops.py
@@ -49,102 +49,103 @@ Operations
 Code details
 ~~~~~~~~~~~~
 """
-from pennylane.operation import Operation
+# TODO[CUSTOM OPS]: Uncomment and replace with Cirq-specific ops
+# from pennylane.operation import Operation
 
 
-class S(Operation):
-    r"""S(wires)
-    S gate.
+# class S(Operation):
+#     r"""S(wires)
+#     S gate.
 
-    .. math:: S = \begin{bmatrix} 1 & 0 \\ 0 & i \end{bmatrix}
+#     .. math:: S = \begin{bmatrix} 1 & 0 \\ 0 & i \end{bmatrix}
 
-    **Details:**
+#     **Details:**
 
-    * Number of wires: 1
-    * Number of parameters: 0
+#     * Number of wires: 1
+#     * Number of parameters: 0
 
-    Args:
-        wires (int): the subsystem the gate acts on
-    """
-    num_params = 0
-    num_wires = 1
-    par_domain = None
-
-
-class T(Operation):
-    r"""T(wires)
-    T gate.
-
-    .. math:: T = \begin{bmatrix}1&0\\0&e^{i \pi / 4}\end{bmatrix}
-
-    **Details:**
-
-    * Number of wires: 1
-    * Number of parameters: 0
-
-    Args:
-        wires (int): the subsystem the gate acts on
-    """
-    num_params = 0
-    num_wires = 1
-    par_domain = None
+#     Args:
+#         wires (int): the subsystem the gate acts on
+#     """
+#     num_params = 0
+#     num_wires = 1
+#     par_domain = None
 
 
-class CCNOT(Operation):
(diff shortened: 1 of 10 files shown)


pennylane_cirq/ops.py
```diff
@@ -49,102 +49,103 @@
...
-    """
-    num_params = 0
-    num_wires = 3
-    par_domain = None
+#     Args:
+#         wires (int): the subsystem the gate acts on
+#     """
+#     num_params = 0
+#     num_wires = 3
+#     par_domain = None
```
josh146: How come this is commented?

pennylane_cirq/simulator_device.py
```diff
@@ -187,7 +187,7 @@ def pre_measure(self):
 
     def probability(self):
         if self.state is None:
-            return None
+            raise qml.DeviceError("Probability can not be computed because the internal state is None.")
```
josh146: :laughing:

tests/test_cirq_interface.py
```diff
@@ -66,6 +66,19 @@ def test_apply(self):
...
+        that was not parametrized before."""
+        operation = CirqOperation(
+            lambda a, b, c: [cirq.X, cirq.Ry(a), cirq.Rx(b), cirq.Z, cirq.Rz(c)]
+        )
+        qubit = cirq.LineQubit(1)
+
+        with pytest.raises(
+            qml.DeviceError, match="CirqOperation must be parametrized before it can be applied."
+        ):
+            operation.apply(qubit)
```
josh146: Maybe I'm just missing context, but I'm a bit confused what this is testing?

tests/test_simulator_device.py
```diff
@@ -529,3 +605,58 @@ def circuit():
...
+class TestSample:
+    """Test sampling."""
+
+    def test_probability_sum_error(self, simulator_device_1_wire):
+        """Test that an error is raised if the probabilites in the 
+        internal state do not sum up to 1."""
+
+        simulator_device_1_wire.state = np.array([1, 1])
+        with pytest.raises(ValueError, match="Probabilites in sampling must sum up to 1."):
+            simulator_device_1_wire.sample("PauliZ", wires=[0], par=[])
```
josh146: Would this be caused by the Cirq simulator itself? :thinking:

josh146: As in, this is never something that could be triggered by the user? (is it worth catching, in that case?)

pennylane_cirq/ops.py
johannesjmeyer: Those were the extra-ops from the plugin-template. I wanted to keep them as a template for when I add additional operations for Cirq but if they are not commented the coverage goes down.

tests/test_cirq_interface.py
johannesjmeyer: The `CirqOperation` class encodes my mapping between Cirq operations and PennyLane gates. For example, `Rot` maps to 
```python
"Rot": CirqOperation(lambda a, b, c: [cirq.Rz(a), cirq.Ry(b), cirq.Rz(c)])
```
which means the successive application of `Rz`, `Ry` and `Rz`. This also needs to be parametrized with the parameters of `Rot`. 

The test checks that an error is raised if a non-parametrized operation is applied to a qubit.

tests/test_simulator_device.py
johannesjmeyer: If you look at the lines below, then you see that I have to normalize the probability vector for numpy's sampling to work. This is caused by small deviations from unity in the Cirq simulator (and probably my postprocessing). As this could possibly hide errors I decided to check that it's not too far off and to throw an error if that was the case.

Do you think this could come from the conversion to a numpy array with `float` dtype? And do you think it is better not to check for an error here?

johannesjmeyer: Okay, after thinking about it I think you're right. We should trust the Cirq simulator to not give us crap.

PR #19, opened 2020-02-03
Port PennyLane-Cirq to QubitDevice

johannesjmeyer:
PennyLane introduces the `QubitDevice` baseclass for plugins in the next release (0.8.0). As `QubitDevice` eases the implementation drastically, this PR ports the PennyLane-Cirq plugin to this base device class.

As a side effect, PennyLane-Cirq now supports tensor observables.

This PR also adresses some changes in Cirq that resulted in depreceation warnings.


pennylane_cirq/cirq_device.py
```diff
@@ -126,18 +116,18 @@ def __init__(self, wires, shots, qubits=None):
...
-        "RX": CirqOperation(lambda phi: cirq.Rx(phi)),
-        "RY": CirqOperation(lambda phi: cirq.Ry(phi)),
-        "RZ": CirqOperation(lambda phi: cirq.Rz(phi)),
-        "Rot": CirqOperation(lambda a, b, c: [cirq.Rz(a), cirq.Ry(b), cirq.Rz(c)]),
-        "CRX": CirqOperation(lambda phi: cirq.ControlledGate(cirq.Rx(phi))),
-        "CRY": CirqOperation(lambda phi: cirq.ControlledGate(cirq.Ry(phi))),
-        "CRZ": CirqOperation(lambda phi: cirq.ControlledGate(cirq.Rz(phi))),
+        "RX": CirqOperation(cirq.rx),
+        "RY": CirqOperation(cirq.ry),
+        "RZ": CirqOperation(cirq.rz),
```
antalszava: I was wondering about these three, how come the codecov had concerns only about these?

pennylane_cirq/cirq_device.py
```diff
@@ -167,58 +159,56 @@ def operations(self):
...
-                    U = self._eigs[Hkey]["eigvec"]
-                else:
-                    # store the eigenvalues corresponding to H
-                    # in a dictionary, so that they do not need to
-                    # be calculated later
-                    w, U = np.linalg.eigh(Hmat)
-                    self._eigs[Hkey] = {"eigval": w, "eigvec": U}
+        # TODO: get pre rotated state here
+
+        # TODO: Remove duplicate code
```
antalszava: I guess this is a comment for creating a new function something like
```python
def apply_operation(self, operation):
    cirq_operation = self._complete_operation_map[operation.name]

    # If command is None do nothing
    if cirq_operation:
        cirq_operation.parametrize(*operation.parameters)

        self.circuit.append(
            cirq_operation.apply(*[self.qubits[wire] for wire in operation.wires])
        )
```
That could be used starting from line 187 and 204

pennylane_cirq/cirq_device.py
```diff
@@ -167,58 +159,56 @@ def operations(self):
...
-                if Hkey in self._eigs:
-                    # retrieve eigenvectors
-                    U = self._eigs[Hkey]["eigvec"]
-                else:
-                    # store the eigenvalues corresponding to H
-                    # in a dictionary, so that they do not need to
-                    # be calculated later
-                    w, U = np.linalg.eigh(Hmat)
-                    self._eigs[Hkey] = {"eigval": w, "eigvec": U}
+        # TODO: get pre rotated state here
```
antalszava: ?

pennylane_cirq/simulator_device.py
```diff
@@ -71,225 +59,115 @@ class SimulatorDevice(CirqDevice):
...
     def __init__(self, wires, shots=1000, analytic=True, qubits=None):
-        super().__init__(wires, shots, qubits)
+        super().__init__(wires, shots, analytic, qubits)
 
-        self.initial_state = None
-        self.simulator = cirq.Simulator()
-        self.result = None
-        self.measurements = None
-        self.state = None
-        self.analytic = analytic
```
antalszava: Here by removing this attribute, it will be created implicitly, right? (Just checking if this was intended).

pennylane_cirq/simulator_device.py
```diff
@@ -71,225 +59,115 @@ class SimulatorDevice(CirqDevice):
...
-            self.initial_state = state_vector
-
-        if self._first_apply:
-            self._first_apply = False
+        self._initial_state = state_vector
 
-    def pre_measure(self):
-        super().pre_measure()
+    def apply(self, operations, rotations=None, **kwargs):
+        super().apply(operations, rotations, **kwargs)
```
antalszava:
```suggestion
        super().apply(operations, rotations=rotations, **kwargs)
```

antalszava: This would be in case `kwargs=={}` is `True` and `apply` being called only with passing `operations` to it, since then I think `rotations` would default to `None` and there might be an error arising from passing too many positional arguments (`None` is going to be passed as the second argument). Might be wrong though.

pennylane_cirq/simulator_device.py
```diff
@@ -71,225 +59,115 @@ class SimulatorDevice(CirqDevice):
...
-                self.marginal_probability(wires).values(), dtype=np.float
-            )
-
-            return (
-                np.dot(eigenvalues ** 2, marginal_probability)
-                - np.dot(eigenvalues, marginal_probability) ** 2
-            )
-        else:
-            return self.sample(observable, wires, par).var()
+            return super().generate_samples()
```
antalszava: How about the case of `analytic==True` and returning `qml.sample`?

tests/test_cirq_device.py
```diff
@@ -32,15 +32,16 @@ class TestCirqDeviceInit:
     def test_default_init(self):
         """Tests that the device is properly initialized."""
 
-        dev = CirqDevice(3, 100)
+        dev = CirqDevice(3, 100, False)
```
antalszava: These test cases could be parameterized as
```python
@pytest.mark.parametrize("analytic", [False, True])
```
and then check both cases.

tests/test_cirq_device.py
```diff
@@ -142,217 +143,104 @@ def test_pre_apply(self, cirq_device_1_wire):
...
-            ("PhaseShift", [2], [cirq.ZPowGate(exponent=2 / np.pi)]),
-            ("PhaseShift.inv", [1.4], [cirq.ZPowGate(exponent=-1.4 / np.pi)]),
-            ("PhaseShift.inv", [-1.2], [cirq.ZPowGate(exponent=1.2 / np.pi)]),
-            ("PhaseShift.inv", [2], [cirq.ZPowGate(exponent=-2 / np.pi)]),
-            ("RX", [1.4], [cirq.Rx(1.4)]),
-            ("RX", [-1.2], [cirq.Rx(-1.2)]),
-            ("RX", [2], [cirq.Rx(2)]),
-            ("RX.inv", [1.4], [cirq.Rx(-1.4)]),
-            ("RX.inv", [-1.2], [cirq.Rx(1.2)]),
-            ("RX.inv", [2], [cirq.Rx(-2)]),
```
antalszava: Perhaps keeping 1-1 rotations tests (e.g. RX, Rot and their inverses) could be nice

pennylane_cirq/cirq_device.py
```diff
@@ -34,21 +34,21 @@
...
 import pennylane as qml
-from pennylane import Device
+from pennylane import QubitDevice
 from pennylane.operation import Operation
 
 from ._version import __version__
-from .cirq_interface import CirqOperation, unitary_matrix_gate
+from .cirq_interface import CirqOperation
 
+# pylint: disable=missing-function-docstring
```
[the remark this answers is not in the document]
johannesjmeyer: Pylint complained about the missing docstrings in the Device class, but those are fine as we don't put them for subclasses of `Device` because they are inherited, no?

Now I realize that this might not work here because this is separate from PL. I should copy the docstrings, shouldn't I?

pennylane_cirq/cirq_device.py
johannesjmeyer: Because here he could be sure that the lambda was not necessary. For example in `lambda: cirq.SWAP`, `cirq.SWAP` is not callable, but `CirqOperation` expects a callable as argument, so the lambda is necessary.

pennylane_cirq/cirq_device.py
```diff
@@ -167,58 +159,56 @@ def operations(self):
...
-        operation = self._complete_operation_map[operation]
-
-        # If command is None do nothing
-        if operation:
-            operation.parametrize(*par)
+    def apply_basis_state(self, basis_state_operation):
+        pass
 
-            self.circuit.append(operation.apply(*[self.qubits[wire] for wire in wires]))
+    def apply_qubit_state_vector(self, qubit_state_vector_operation):
```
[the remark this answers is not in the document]
johannesjmeyer: Those are totally needed and used in the `apply` method. Those methods are necessary for me to inject the state preparations into the apply routine in the subclass.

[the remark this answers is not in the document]
johannesjmeyer: I wanted to avoid to reimplement `apply` in the subclass basically.

pennylane_cirq/cirq_device.py
johannesjmeyer: Exactly! I implemented it

pennylane_cirq/simulator_device.py
johannesjmeyer: You mean `analytic`? It is now created in the superclass.

pennylane_cirq/simulator_device.py
```diff
@@ -71,225 +59,115 @@ class SimulatorDevice(CirqDevice):
...
-
-            if not self.analytic:
-                raise qml.DeviceError(
-                    "The operation BasisState is only supported in analytic mode."
-                )
+    def apply_basis_state(self, basis_state_operation):
+        if not self.analytic:
+            raise qml.DeviceError(
+                "The operation BasisState is only supported in analytic mode."
+            )
```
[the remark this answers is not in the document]
johannesjmeyer: The other device is the superclass to this device and is intended as an abstract base class. Maybe I should make this more explicit. The explanation why this method is necessary is given above.

pennylane_cirq/cirq_device.py
antalszava: Hmm, that's a fair point, thanks! :)

pennylane_cirq/simulator_device.py
antalszava: Yep! Cool!

pennylane_cirq/cirq_device.py
johannesjmeyer: Actually I'm a little confused with the handling of pre-rotation state and rotated state. In the implementation of `default.qubit`, the probability is given for the rotated state but the state is given for the pre_rotated_state. I think this is somewhat inconsistent, isn't it?

Here in Cirq I have the additional problem that I can't really extract the pre_rotated_state without much hassle

johannesjmeyer: What is your opinion on this, @antalszava ?

pennylane_cirq/simulator_device.py
johannesjmeyer: I removed the argument `rotations` entirely now to be consistent with the signature of `apply` from the `QubitDevice` class.

pennylane_cirq/simulator_device.py
johannesjmeyer: I see no problem with that? The logic of deciding when to call `generate_samples` is implemented in `QubitDevice`, so I just have to assure that samples are properly generated when it is called. And for analytic mode I can safely piggyback onto the default implementation.

tests/test_cirq_device.py
johannesjmeyer: They are still there, it is just misaligned due to the blacking

pennylane_cirq/cirq_device.py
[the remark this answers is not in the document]
josh146: Oh, that's odd. When you build the docs, are the inherited docstrings inserted? I wouldn't think you should need to re-write the docstring.

pennylane_cirq/cirq_device.py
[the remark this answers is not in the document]
josh146: Lol based on the emojis this is my most controversial GitHub comment :laughing:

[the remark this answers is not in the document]
johannesjmeyer: I hope that the development team can sustain such an astonishing level of controversy 🤣

pennylane_cirq/simulator_device.py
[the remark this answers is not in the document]
johannesjmeyer: I made `CirqDevice` abstract now

pennylane_cirq/cirq_device.py
johannesjmeyer: And @josh146

pennylane_cirq/cirq_device.py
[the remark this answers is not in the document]
johannesjmeyer: It works in the docs 👍

pennylane_cirq/simulator_device.py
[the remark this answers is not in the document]
josh146: That definitely makes it clearer!

pennylane_cirq/cirq_device.py
josh146: This is slightly intentional. Now that there is `qml.prob`, we no longer want the user to access `dev.probability`, so for consistency this should always represent the actual device probability.

For state, however, we are still in an awkward position where there is no equivalent to extract it from the qnode yet

johannesjmeyer: So what do you think of me keeping the post-rotation state as `dev.state`? It is possible to extract the pre-rotation state, but it will not make the code clearer...

josh146: I think it is fine for now. `dev.state` should only ever be used for debugging, and we can attach caveats to it that depend on the device (e.g. a lot of hardware simulators don't even store any state)

johannesjmeyer: I added a docstring that explains this

pennylane_cirq/simulator_device.py
antalszava: This is the part of the `device.execute()`:
```python
        if (not self.analytic) or circuit.is_sampled:
            self._samples = self.generate_samples()
```
for `self.analytic==True and circuit.is_sampled` `SimulatorDevice.generate_samples()` will be called and since `self.analytic==True` the if statement evaluates to `True` on line 158, and `QubitDevice.generate_samples()` will be called.

Whereas, if `self.analytic==False and circuit.is_sampled`, `SimulatorDevice.generate_samples()` still gets called, but the if statement on line 158 evaluates to `False` and the code block below gets executed.

If this is so (and I didn't confuse myself or mess up something here :sweat_smile: ), then how come that there is a different logic for these two cases?

johannesjmeyer: All you say is totally correct. There is a different logic for the `analytic==False` case because in that case I don't generate the samples myself but I use Cirq's own measurement scheme. I did this because I wanted to use as much of Cirq's inbuilt API as possible

johannesjmeyer: This is the difference between `simulator.run` (`analytic==False`) and `simulator.simulate` (`analytic==True`)

requirements.txt
```diff
@@ -1,3 +1,3 @@
-pennylane>=0.7
+pennylane>=0.8
```
antalszava: This was extending the functionality of #21 with `QubitDevice` being used in this plugin.

PR #20, opened 2020-02-04
Include test case for Cirq v0.6 and v0.7 as well

antalszava:
**Context**
A test case in `test_apply` output different messages with Cirq version `0.6` and `0.7`.

**Description of Change**
This PR creates fixtures for these two cases, such that the version number of the Cirq package can remain as is in the `requirements.txt` file.

It basically includes both the old and new test message from #18 and skips exactly one of them based on the Cirq version.

**Benefits**
This way, the test suite will pass with both `cirq==0.6` and `cirq>=0.7`.

**Drawbacks**
Branching off for a test case might be unfavourable.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 import numpy as np
 import pytest
+from packaging import version
 
+from cirq import __version__
 from pennylane_cirq import SimulatorDevice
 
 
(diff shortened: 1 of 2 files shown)


josh146:
Nice catch @antalszava! As with the Qiskit case, I would also suggest here supporting only the latest cirq version (0.7). So far, with Cirq so new, there are no use-cases for supporting old versions (this might change in the future, once there is stability in the API/Cirq provides ongoing support for older versions)

antalszava:
Cool, thanks! :) I'll close this one!

PR #24, opened 2020-04-07
Rename probability to analytic_probability

antalszava:
This change follows-up the change in the structure of `QubitDevice`:
https://github.com/XanaduAI/pennylane/pull/573


antalszava:
Thanks! It was the `self._state is None` case, added a small test for coverage on that! :blush:

PR #28, opened 2020-05-15
Update requirements

thisac:
* Python 3.5 support is removed (since Cirq doesn't officially support it).
* PennyLane requirements are bumped up to 0.9.
* Cirq requirements are bumped to 0.7 in `README.rst` (was already the case in `requirements.txt`).
* Minor changes are made in `README.rst` to work properly on PyPI (similiar to what was done [here](https://github.com/XanaduAI/pennylane-sf/pull/35)).

**Note**: Travis will fail because of using `networkx==2.3` instead of `networkx==2.4`, as written in `requirements.txt`. This is fixed in #27, and should thus work as soon as that PR is merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,6 @@ classifiers = [
     "Programming Language :: Python",
     # Make sure to specify here the versions of Python supported
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.5",
     "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
     "Programming Language :: Python :: 3 :: Only",
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ dist: xenial
 python:
   - 3.6
   - 3.7
+  - 3.8
 install:
   - pip install pip --upgrade
   - pip install -e git+https://github.com/XanaduAI/pennylane.git#egg=pennylane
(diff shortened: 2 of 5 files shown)


josh146:
Don't forget to also update the changelog to change `0.9.0-dev` -> `0.9.0` after #27 is merged but before this is merged :)

thisac:
> Don't forget to also update the changelog to change `0.9.0-dev` -> `0.9.0` after #27 is merged but before this is merged :)

I thought I would do a second "Incrementing the version number to 0.9.0" with the release changes before uploading to PyPi. Should I just do it here instead?

josh146:
> I thought I would do a second "Incrementing the version number to 0.9.0" with the release changes before uploading to PyPi. Should I just do it here instead?

Sure! Wasn't sure if it was meant for this PR or not.

setup.py
```diff
@@ -56,7 +56,6 @@
     "Programming Language :: Python",
     # Make sure to specify here the versions of Python supported
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.5",
     "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
```
antalszava, line 60:
```suggestion
    "Programming Language :: Python :: 3.7",
    "Programming Language :: Python :: 3.8",
```

PR #33, opened 2020-07-09
Update branch

co9olguy:
@josh146 I realized that the following branch is the active development branch:
https://github.com/lhenriet/Cirq/tree/pasqal_new

(not sure if I got the formatting right in `requirements.txt`)

You'll note that this branch also specifies `ThreeDGridQubit`s using floats

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pennylane>=0.9
 google-api-core[grpc]<=1.14.0
-git+git://github.com/lhenriet/Cirq/@pasqal_new
+git+https://github.com/lhenriet/Cirq.git@pasqal_new
 numpy~=1.16
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 pennylane>=0.9
 google-api-core[grpc]<=1.14.0
-git+https://github.com/lhenriet/Cirq.git
+git+https://github.com/lhenriet/Cirq/tree/pasqal_new
 numpy~=1.16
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
 pennylane>=0.9
 google-api-core[grpc]<=1.14.0
-git+https://github.com/lhenriet/Cirq.git
+git+git://github.com/lhenriet/Cirq/@pasqal_new
```
josh146, line 3:
```suggestion
git+https://github.com/lhenriet/Cirq.git@pasqal_new
```

josh146:
Nice catch!

It looks like there have been big changes to the device: `PasqalDevice` has been split into `PasqalDevice` (which only accepts named qubits unrelated to 3D positions) and `PasqalVirtualDevice`, which accepts `ThreeDGridQubit`, `LineQubit`, or `GridQubit`.

josh146:
@co9olguy got the tests passing by simply renaming `ThreeDGridQubit` to `ThreeDQubit` everywhere, but this requires further investigation to make sure that we are using the updated qubits/pasqal device properly.

co9olguy:
I think we'll have to keep an eye on any changes coming from the branch, but otherwise it shouldn't affect our ability to make a demo

josh146:
Would we need to change the PL device to instead use `PasqalVirtualDevice` however?

co9olguy:
Don't know. If the tests are passing, I don't think we need it. But we can play it by ear as we create a demo

PR #36, opened 2020-08-04
Add QSim device

thisac:
Adds a new [QSim device](https://github.com/quantumlib/qsim/blob/master/docs/cirq_interface.md) by importing `qsimcirq` and creating a QSim circuit using a Cirq circuit as `qsimcirq.QSimCircuit(cirq_circuit=cirq.Circuit())` and then using `qsimcirq.QSimSimulator`. 

It can be installed with `pip install qsimcirq`.


thisac:
Quick benchmark for the following circuit:

```python
def circuit(x, y, z):
    for j in range(i-1):
        qml.RZ(z, wires=[j])
        qml.RY(y, wires=[j])
        qml.RX(x, wires=[j])
        qml.CNOT(wires=[j, j+1])

    return qml.expval(qml.PauliZ(wires=i-1))
```

image

co9olguy:
Nice @thisac, good to see this making progress!

josh146:
Awesome @thisac! 🎉 

> Note: this branch currently imports qsimcirq and thus assumes that it is installed. We probably either want to have some sort of check whether the user has qsimcirq installed or add it to the requirements.

I had a look at the qsim PyPI package, and it looks like it includes platform-specific code (C++), but binary wheels are not provided. Since it could be hard for users to install, I suggest not adding it to `setup.py`, but taking the same approach we do with optional dependencies in PL --- use a `try`/`except` around the import, and make sure the qsim import doesn't occur on other device loads!

pennylane_cirq/simulator_device.py
```diff
@@ -132,7 +132,8 @@ def apply(self, operations, **kwargs):
 
         # We apply an identity gate to all wires, otherwise Cirq would ignore
         # wires that are not acted upon
-        self.circuit.append(cirq.IdentityGate(len(self.qubits))(*self.qubits))
+        for qb in self.qubits:
+            self.circuit.append(cirq.IdentityGate(1)(qb))
```
trbromley, line 136: So in the new approach we're looping over individual qubits, but before we passed all of the qubits to a single identity gate?

Was this done for clarity or was it broken?

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+
+try:
+    import qsimcirq
+except ImportError:
+    raise ImportError(
+        "QSim Cirq is needed for the QSim devices to work."
+        "\nIt can be installed using pip:"
+        "\n\npip install qsimcirq"
+    )
```
trbromley, line 28: Are we avoiding making this a dependency as `qsimcirq` might be difficult to install?

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+
+from .simulator_device import SimulatorDevice
+from .cirq_device import CirqDevice
+
+
+class QSimDevice(SimulatorDevice):
+    r"""QSim device for PennyLane.
+    Args:
```
trbromley, line 36:
```suggestion

    Args:
```

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+    short_name = "cirq.qsim"
+
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None):
+        super().__init__(wires, shots, analytic, qubits)
+        self.circuit = qsimcirq.QSimCircuit(cirq_circuit=cirq.Circuit())
+        self._simulator = qsimcirq.QSimSimulator()
+
+        self._initial_state = None
+        self._result = None
+        self._state = None
```
trbromley, line 58: Do we need these, since they are initialized in `QSimDevice.super()`?

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class QSimhDevice(SimulatorDevice):
+    r"""QSimh device for PennyLane.
+    Args:
```
trbromley, line 63:
```suggestion

    Args:
```

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+                'w': 0,
+                'p': 0,
+                'r': 1
+            }
+        self.circuit = cirq.Circuit()
+        self._simulator = qsimcirq.QSimhSimulator(qsimh_options)
+
+        self._initial_state = None
+        self._result = None
+        self._state = None
```
trbromley, line 95: Similarly, maybe we don't need these

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+            to be >= 1. In analytic mode, shots indicates the number of entries
+            that are returned by ``device.sample``.
+        analytic (bool): indicates whether expectation values and variances should
+            be calculated analytically
+        qubits (List[cirq.Qubit]): A list of Cirq qubits that are used
+            as wires. The wire number corresponds to the index in the list.
+            By default, an array of ``cirq.LineQubit`` instances is created.
+        qsimh_options (dict): A dictionary with options for the QSimh simulator. See the `QSim
+            usage documentation <https://github.com/quantumlib/qsim/blob/master/docs/usage.md>`__
+            for further details.
```
trbromley, line 76: This is probably the plan already, but would be good to give a nice explanation of this in the docs.

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+    short_name = "cirq.qsimh"
+
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None, qsimh_options=None):
+        super().__init__(wires, shots, analytic, qubits)
+        if qsimh_options is None:
+            qsimh_options = {
+                'k': [0],
+                'w': 0,
+                'p': 0,
+                'r': 1
```
trbromley, line 88: Are these options required or could we use a qsimh default?

trbromley: Also, would it make sense to make `qsimh_options` a required argument in this device? I'm not sure why those numbers above are good, so feel it might be good to force the user to decide.

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,111 @@
...
+        for qb in self.qubits:
+            self.circuit.append(cirq.IdentityGate(1)(qb))
+
+        if self.analytic:
+            state = self._simulator.compute_amplitudes(
+                program=self.circuit,
+                bitstrings=list(range(2 ** len(self.wires)))
+            )
+
+            self._state = np.array(state)
```
trbromley, line 111: Things piece together nicely!

setup.py
```diff
@@ -35,7 +35,10 @@
     "license": "Apache License 2.0",
     "packages": ["pennylane_cirq"],
     "entry_points": {"pennylane.plugins": ["cirq.simulator = pennylane_cirq:SimulatorDevice",
-                                           "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice"],},
+                                           "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice",
+                                           "cirq.qsim = pennylane_cirq.qsim_device:QSimDevice",
+                                           "cirq.qsimh = pennylane_cirq.qsim_device:QSimhDevice",
+                                           "cirq.qflex = pennylane_cirq.qflex_device:QFlexDevice"],},
```
trbromley, line 41:
```suggestion
```

trbromley (commented):
Thanks @thisac, looks great so far and also works on a simple prototype I made! I'll be happy to take another look when the tests and supporting docs are added.
:100:

pennylane_cirq/simulator_device.py
thisac: @trbromley It didn't work applying a single identity gate, so this was the simplest work-around I could think of. I noticed that it's actually done for the `SimulatorDevice` as well during the wires refactor, but I'm not sure about that change exactly.

pennylane_cirq/qsim_device.py
thisac: Yes, I think that was the case. I didn't have any issues myself, but I believe @josh146 mentioned something about this being potentially difficult to build.

pennylane_cirq/qsim_device.py
thisac: Oh, you're right. Removed them now. Thanks @trbromley! :tada:

pennylane_cirq/qsim_device.py
thisac: I'm quite sure that these options are required. I just put in the default recommended ones from the website here.

I guess having the options as a required argument makes a lot of sense (since they are required by qsimh). :thinking:

tests/test_qsim_device.py
```diff
@@ -0,0 +1,382 @@
+# Copyright 2018 Xanadu Quantum Technologies Inc.
```
thisac, line 1: The tests in this file have been copied from `test_cirq_device.py`, with only some minor modifications to work with `QSimDevice`.

tests/test_qsimh_device.py
```diff
@@ -0,0 +1,400 @@
+# Copyright 2018 Xanadu Quantum Technologies Inc.
```
thisac, line 1: Same with these tests. They have been copied from `test_cirq_device.py`, with only some minor modifications to work with `QSimhDevice`. In addition, two tests for the `qsimh_options` kwarg has been added.

thisac:
Sorry for taking so long to finish this. I've done some minor changes and added tests (mainly copied from the CirqDevice tests, with some minor alterations). Two things to note is that

1. qsimcirq doesn't support `ControlledGate` (i.e. gates constructed using the `controlled_by()` method), so those tests have simply been removed from the test-files for now. I might look into it tomorrow and see if any more tests can be added to replace these.

2. There seems to be an issue with `cirq.inverse(gate)` for gates `cirq.X`, `cirq.Y` and `cirq.Z` where it sets the exponent to an integer (-1), while qsim only assumes/checks for float exponents. Thus `PauliX.inv()`, `PauliY.inv()`, `PauliZ.inv()` raise fairly obscure C++ errors when used with qsim and qsimh.

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,110 @@
...
+            usage documentation <https://github.com/quantumlib/qsim/blob/master/docs/usage.md>`__
+            for further details.
+    """
+    name = "QSimh device for PennyLane"
+    short_name = "cirq.qsimh"
+
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None, qsimh_options=None):
+        super().__init__(wires, shots, analytic, qubits)
+        if qsimh_options is None:
+            qsimh_options = {"k": [0], "w": 0, "p": 0, "r": 1}
```
josh146, line 91: The qsimh options are quite opaque 😆 You don't often see so many single letter dictionary keys

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,110 @@
...
+        self.circuit = None
+        self.qsimh_options = qsimh_options
+        self._simulator = qsimcirq.QSimhSimulator(qsimh_options)
+
+    def apply(self, operations, **kwargs):
+        CirqDevice.apply(self, operations, **kwargs)
+
+        # TODO: remove the need for this hack by keeping better track of unused wires
+        # We apply identity gates to all wires, otherwise Cirq would ignore
+        # wires that are not acted upon
```
josh146, line 101: Is this todo copied from the existing Cirq device? Looks familiar

pennylane_cirq/simulator_device.py
```diff
@@ -135,6 +135,7 @@ def apply(self, operations, **kwargs):
         # TODO: remove the need for this hack by keeping better track of unused wires
```
josh146, line 135: I suppose this answers my question above!

setup.py
```diff
@@ -35,7 +35,9 @@
     "license": "Apache License 2.0",
     "packages": ["pennylane_cirq"],
     "entry_points": {"pennylane.plugins": ["cirq.simulator = pennylane_cirq:SimulatorDevice",
-                                           "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice"],},
+                                           "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice",
+                                           "cirq.qsim = pennylane_cirq.qsim_device:QSimDevice",
+                                           "cirq.qsimh = pennylane_cirq.qsim_device:QSimhDevice"],},
```
josh146, line 40: 😍

josh146 (commented):
Looks good @thisac! Will be awesome to have this merged in 💪 

The final part of this PR would be to add documentation for the two new devices (i.e., by making `doc/devices/qsim.rst` and `qsimh.rst`, adding cards to `index.rst`, and adding the new pages to the TOC).

This would be the place to discuss:

* The reasons to pick qsim/qsimh over the Cirq simulator (could even include the benchmarking plot here!)

* Any additional installation instructions

* Minimal code examples

* Details on any device options (I think relevant for qsimh --- doesn't have to be too detailed, and can link back to the qsim docs)

* Any 'gotchas' the user should be aware of

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,110 @@
...
+        qubits (List[cirq.Qubit]): A list of Cirq qubits that are used
+            as wires. The wire number corresponds to the index in the list.
+            By default, an array of ``cirq.LineQubit`` instances is created.
+    """
+    name = "QSim device for PennyLane"
+    short_name = "cirq.qsim"
+
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None):
+        super().__init__(wires, shots, analytic, qubits)
+        self.circuit = None
```
trbromley, line 56: For my understanding, why not `self.circuit = qsimcirq.QSimCircuit(cirq_circuit=self.circuit)` here? I guess it makes more sense in `reset()`?

pennylane_cirq/qsim_device.py
trbromley: Yes, I would vote to make `qsimh_options` a required arg, unless we're quite confident that those options make sense as a default.

pennylane_cirq/qsim_device.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None, qsimh_options=None):
+        super().__init__(wires, shots, analytic, qubits)
+        if qsimh_options is None:
+            qsimh_options = {"k": [0], "w": 0, "p": 0, "r": 1}
+        self.circuit = None
+        self.qsimh_options = qsimh_options
+        self._simulator = qsimcirq.QSimhSimulator(qsimh_options)
+
+    def apply(self, operations, **kwargs):
+        CirqDevice.apply(self, operations, **kwargs)
```
trbromley, line 97: Just for my understanding: there is something in the `SimulatorDevice` version of `apply()` that we don't want to use, preventing us from calling `super().apply()` here?

Would it make sense for this device to inherit directly from `CirqDevice`? (maybe not, perhaps `SimulatorDevice` is useful in other places).

tests/test_qsim_device.py
```diff
@@ -0,0 +1,382 @@
...
+import qsimcirq
+import pennylane as qml
+from pennylane.wires import Wires
+import pytest
+import numpy as np
+
+from pennylane_cirq.qsim_device import QSimDevice
+
+
+@patch.multiple(QSimDevice, __abstractmethods__=set())
```
trbromley, line 30: Never seen that before!!

trbromley: What does it do :laughing:

tests/test_qsim_device.py
```diff
@@ -0,0 +1,382 @@
...
+        qubits = [
+            cirq.GridQubit(0, 1),
+            cirq.GridQubit(1, 0),
+            cirq.GridQubit(0, 0),
+            cirq.GridQubit(1, 1),
+        ]
+
+        dev = QSimDevice(4, 100, False, qubits=qubits)
+        assert len(dev.qubits) == 4
+        assert dev.qubits == sorted(qubits)
```
trbromley, line 82: Nice!

tests/test_qsim_device.py
```diff
@@ -0,0 +1,382 @@
...
+        ]
+
+        with pytest.raises(
+            qml.DeviceError,
+            match="The number of given qubits and the specified number of wires have to match",
+        ):
+            dev = QSimDevice(3, 100, False, qubits=qubits)
+
+
+class TestQSimDeviceIntegration:
```
trbromley, line 101: Are these integration tests?

tests/test_qsim_device.py
```diff
@@ -0,0 +1,382 @@
...
+    )
+    def test_apply_single_wire(self, qsim_device_1_wire, gate, expected_cirq_gates):
+        """Tests that apply adds the correct gates to the circuit for single-qubit gates."""
+
+        qsim_device_1_wire.reset()
+
+        # try:
+        qsim_device_1_wire.apply([gate])
+        # except:
+            # sdfsd
```
trbromley, line 293:
```suggestion
        qsim_device_1_wire.apply([gate])
```

tests/test_qsimh_device.py
```diff
@@ -0,0 +1,400 @@
...
+import cirq
+import pennylane as qml
+from pennylane.wires import Wires
+import pytest
+import numpy as np
+
+from pennylane_cirq.qsim_device import QSimhDevice
+
+
+@patch.multiple(QSimhDevice, __abstractmethods__=set())
```
trbromley, line 29: Unsure if we need this also here, isn't it just to help with `CirqDevice` which has some abstract methods. Do we still have abstract methods in `QSimhDevice`?

tests/test_qsimh_device.py
```diff
@@ -0,0 +1,400 @@
...
+                'w': 2,
+                'p': 4,
+                'r': 2
+            }
+        dev = QSimhDevice(3, 100, False, qsimh_options=qsimh_options)
+
+        assert qsimh_options.items() <= dev._simulator.qsimh_options.items()
+
+
+class TestQSimhDeviceIntegration:
```
trbromley, line 118: Also here, they still seem more like unit tests than integration :thinking:

tests/test_qsimh_device.py
```diff
@@ -0,0 +1,400 @@
...
+    )
+    def test_apply_single_wire(self, qsimh_device_1_wire, gate, expected_cirq_gates):
+        """Tests that apply adds the correct gates to the circuit for single-qubit gates."""
+
+        qsimh_device_1_wire.reset()
+
+        # try:
+        qsimh_device_1_wire.apply([gate])
+        # except:
+            # sdfsd
```
trbromley, line 310:
```suggestion
        qsimh_device_1_wire.apply([gate])
```

trbromley (commented):
Thanks @thisac! I noticed in the tests that the devices don't support inverses and some controlled gates - we should probably update the capabilities to reflect that (e.g., `"inverse_operations" = True`) and maybe update the `_operation_map` by taking out the relevant entries?

pennylane_cirq/qsim_device.py
thisac: Yes, it was written differently before, but I noticed the `TODO:` in the existing Cirq device, so I thought it might be good to duplicate it here so we don't forget about this one as well (unfortunately this  part needed to be duplicated).

pennylane_cirq/qsim_device.py
thisac: It was there before, but following the logic in the Cirq device, I changed it to be initialized as `None`. I noticed it since there were tests testing for specifically that. Not sure if there's a reason for having it this way, but as long as `reset` is used the way it is now, it should work equally well.

pennylane_cirq/qsim_device.py
thisac: Yes, that's exactly the case. `SimulatorDevice.apply` uses the method `_simulator.simulate` which qsimh doesn't have (qsim has it, and thus simply uses `SimulatorDevice.apply`). 

> Would it make sense for this device to inherit directly from CirqDevice? (maybe not, perhaps SimulatorDevice is useful in other places).

Not really, since quite a lot of the methods in `SimulatorDevice` are needed (e.g. `Cirq.apply` calls several abstract methods that only exist in `SimulatorDevice`). The alternative would be to make qsimh inherit from `CirqDevice` and rewrite many of the functions from `SimulatorDevice`, but I'm not sure if I think that actually makes things better. It might seem more structured, but would mean maintaining two version of many of the functions. I was thinking about this while writing it, and decided on the current solution. :slightly_smiling_face:

tests/test_qsimh_device.py
thisac: Oh, my bad, I copied the tests for `CirqDevice` rather than for `SimulatorDevice`. Thanks for catching this @trbromley! :facepalm:

tests/test_qsimh_device.py
thisac: Another :facepalm:. I find this to be an easy way to get the tests to stop and enter the terminal for debugging. There's probably a better way, but this generally works quite nicely. :laughing:

pennylane_cirq/qsim_device.py
thisac: @josh146 Would you recommend just keeping the `TODO`s here and simply ignore the pylint complaint for this PR?

pennylane_cirq/qsim_device.py
```diff
@@ -53,13 +54,44 @@ class QSimDevice(SimulatorDevice):
...
     def reset(self):
         # pylint: disable=missing-function-docstring
         super().reset()
-        self.circuit = qsimcirq.QSimCircuit(cirq_circuit=self.circuit)
+        self.circuit = qsimcirq.QSimCircuit(cirq_circuit=cirq.Circuit())
+
+    def _apply_basis_state(self, basis_state_operation):
+        # pylint: disable=missing-function-docstring
+        if not self.analytic:
+            raise qml.DeviceError("The operation BasisState is only supported in analytic mode.")
```
trbromley, line 68: I'm curious why it doesn't work in non-analytic mode?

pennylane_cirq/qsim_device.py
trbromley: Perfect, thanks for the explanation!

pennylane_cirq/qsim_device.py
trbromley: Thanks!

pennylane_cirq/qsim_device.py
thisac: Not sure actually. I copied it over from the SimlatorDevice class. :thinking:

thisac:
Thanks @trbromley  and @josh146 for the reviews. I realized that I'd copied over the wrong tests, so I've fixed that now. While doing that I noticed quite a few things that didn't work. I've added some temporary comments in the tests as well as updated the GitHub comments at the top of each test file, where I've detailed the issues.

I still have to update the docs; look through the tests again and add qsim/qsimh specific tests; make qsimh_options a required kwarg (I thinks that's a good idea, and simply clearing it up in the docs); and, finally, figure out what to do with the failing/commented out tests. Also, I'm not sure if/how qsim can let the integration tests know that it doesn't support `QubitStateVector` while still allowing it for the other Cirq devices. :thinking:

pennylane_cirq/qsim_device.py
josh146: Yep, fine to simply ignore pylint here!

tests/test_qsim_device.py
josh146: This is a way of allowing you to instantiate an abstract class for testing 🙂 Essentially, the `__abstractmethods__` method is temporarily patched to an empty set, so that the class looks to have no abstract methods

pennylane_cirq/qsim_device.py
josh146: (Is this true for qsim? or is it a Cirq restriction?)

josh146: It might be worth removing `BasisState` from the supported operations, so that PL will always default to the decomposition (e.g., PauliX gates), which should always be supported.

josh146:
Thanks @thisac!

> Also, I'm not sure if/how qsim can let the integration tests know that it doesn't support QubitStateVector while still allowing it for the other Cirq devices. 🤔

Can you simply overwrite the `operations` attribute/property in the `QSimDevice` class, rather than relying on the inherited supported operations coming from the parent simulator device?

co9olguy:
I've removed my review request because I don't really have anything specific to offer on top of what @josh146  and @trbromley can already offer for code review.

I'd suggest if any questions/clarification comes up for how to deal with ordering, `LineQubit`s, `GridQubit`s, etc. in PennyLane-Cirq, then definitely reach out to me. I probably would understand best on those issues

doc/index.rst
```diff
@@ -31,7 +31,17 @@ Currently, PennyLane-Cirq provides two Cirq devices for PennyLane:
...
+
+.. devicegalleryitem::
+    :name: 'cirq.qsim'
+    :description: The qsim simulator backend.
+    :link: devices/qsim.html
+
+.. devicegalleryitem::
+    :name: 'cirq.qsimh'
+    :description: The qsimh simulator backend.
+    :link: devices/qsim.html
```
thisac, line 43: Wasn't sure about having two cards link to the same page. I only added a single page (for both qsim and qsimh), since they work very similarly. Do you think it's better having two separate pages, or perhaps just keeping the qsim card here, even though technically qsimh is a separate device? :thinking:

pennylane_cirq/qsim_device.py
```diff
@@ -110,51 +104,38 @@ class QSimhDevice(SimulatorDevice):
...
-    def _apply_qubit_state_vector(self, qubit_state_vector_operation):
+        ops = set(self._operation_map.keys()) - {
+            "QubitStateVector",
+            "BasisState",
+            "CRX",
+            "CRY",
+            "CRZ",
+            "CRot",
+        }
+        return ops
```
josh146, line 129: This is much cleaner!

tests/test_qsim_device.py
```diff
@@ -37,6 +37,51 @@ def test_device_loading(self):
 
         assert isinstance(dev, QSimDevice)
 
+    @pytest.mark.parametrize("analytic", [True, False])
+    @pytest.mark.parametrize("shots", [8192])
+    def test_one_qubit_circuit(self, shots, analytic, tol):
+        """Test that devices provide correct result for a simple circuit"""
```
josh146, line 43: A good test to have 🙂

tests/test_qsim_device.py
```diff
@@ -173,14 +204,17 @@ def test_apply_operation_state_preparation(
     def test_apply_operation_single_wire_with_parameters(
         self, qsim_device_1_wire, tol, op, input, expected_output, par
     ):
-        """Tests that applying an operation yields the expected output state for single wire
+        """Tests that applying an operation yields the expected output probabilities for single wire
```
josh146, line 207: :+1:

josh146 (approved):
Nice work @thisac! I left a couple of suggestions in the documentation; once those are sorted it should be ready to go 🙂

doc/devices/qsim.rst
```diff
@@ -0,0 +1,113 @@
+The qsim device
+===============
+
+qsim is a Schrödinger full state-vector simulator while qsimh is a hybrid
+Schrödinger-Feynman simulator, cutting the qubit lattice into two parts and then
+using qsim to simulate each part individually. For further details see the `qsim
+website <https://github.com/quantumlib/qsim>`__. These simulators are
+considerably faster compared to the Cirq simulator for many qubits, as can be
+seen in the benchmark below.
```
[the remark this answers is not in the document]
thisac: Thanks @josh146. This is much better. :100:

doc/devices/qsim.rst
```diff
@@ -0,0 +1,113 @@
...
+The wire of each qubit corresponds to its index in the ``qubit`` list. In the above example,
+the wire 2 corresponds to ``cirq.GridQubit(1, 0)``.
+
+If no qubits are given, the plugin will create an array of ``LineQubit`` instances.
+
+qsimh options
+^^^^^^^^^^^^^
+
+qsimh requires specific options to be set for the simulator. These can be passed
+by the positional argument ``qsimh_options``. See the `QSimh usage documentation
```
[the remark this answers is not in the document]
thisac: I've been super-confused whether to spell it QSim or qsim (since I've seen both in different places). They use qsim and qsimh on the website and in the supremacy paper, so I went with that. :laughing:

doc/index.rst
thisac: I think that might be the best way to go as well actually. I've updated it now (but kept the line `Currently, PennyLane-Cirq provides four Cirq devices for PennyLane:`, since there are still, technically, four devices).

trbromley (approved):
Thanks @thisac, looks awesome! Left some suggestions on the docs.

doc/devices/qsim.rst
```diff
@@ -0,0 +1,120 @@
...
+
+    @qml.qnode(dev)
+    def circuit(x, y, z):
+        qml.RZ(z, wires=[0])
+        qml.RY(y, wires=[0])
+        qml.RX(x, wires=[0])
+        qml.CNOT(wires=[0, 1])
+        return qml.expval(qml.PauliZ(wires=1))
+
+You can then execute the circuit like any other function to get the quantum mechanical expectation value.
```
thisac, line 45:
```suggestion
You can then execute the circuit like any other function to get the quantum mechanical expectation value:
```

setup.py
```diff
@@ -36,6 +36,8 @@
     "packages": ["pennylane_cirq"],
     "entry_points": {"pennylane.plugins": ["cirq.simulator = pennylane_cirq:SimulatorDevice",
                                            "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice",
+                                           "cirq.qsim = pennylane_cirq:QSimDevice",
+                                           "cirq.qsimh = pennylane_cirq:QSimhDevice"],},
```
thisac, line 40:
```suggestion
                                           "cirq.qsimh = pennylane_cirq:QSimhDevice",

```

setup.py
```diff
@@ -36,6 +36,9 @@
     "packages": ["pennylane_cirq"],
     "entry_points": {"pennylane.plugins": ["cirq.simulator = pennylane_cirq:SimulatorDevice",
                                            "cirq.mixedsimulator = pennylane_cirq:MixedStateSimulatorDevice",
+                                           "cirq.qsim = pennylane_cirq:QSimDevice",
+                                           "cirq.qsimh = pennylane_cirq:QSimhDevice",
+
```
thisac, line 41:
```suggestion
```

josh146:
Thanks @thisac for sorting out the test issue. This looks ready to merge, but before we do so can I get you to update the changelog? You can use the same style as the Pasqal entry in the changelog (and use some of the intro on the `qsim.rst` usage page).

On another note: we should look into using [pytest-randomly](https://pypi.org/project/pytest-randomly/) to avoid this testing issue in future. In particular, pytest-randomly:

* Randomly shuffles the order of test items

* Resets random.seed() at the start of every test case and test to a fixed number

PR #37, opened 2020-08-05
[WIRES] Make plugin compatible with wires refactor

mariaschuld:
This PR updates the plugin to support the new wire management in PennyLane master.


.github/CHANGELOG.md
```diff
@@ -6,6 +6,22 @@
...
+  ``` python
+  
+  dev = qml.device('cirq.simulator', wires=['q1', 'ancilla', 0, 1]
+  
+  def circuit():
+    qml.Hadamard(wires='q1')
+    qml.CNOT(wires=[1, 'ancilla'])
+    ...
+  ```
+  [#XX](https://github.com/PennyLaneAI/pennylane-XX/pull/XX)
```
mariaschuld, line 23:
```suggestion
  [#37](https://github.com/PennyLaneAI/pennylane-cirq/pull/37)
```

mariaschuld:
In future it would be also nice to be a bit cleverer about how to support cirq's GridQubits and LineQubits...But this is a minimal version against the breaking changes!

pennylane_cirq/cirq_device.py
```diff
@@ -71,6 +71,8 @@ def __init__(self, wires, shots, analytic, qubits=None):
 
         self.circuit = None
 
+        device_wires = self.map_wires(self.wires)
```
thisac, line 74: Oh, so if I understand this correctly, `self.map_wires` will in this case go from e.g. `self.wires -> ['q1', 'ancilla', 0, 1]` to `device_wires -> [0, 1, 2, 3]`. And then the list of wires can be extracted from the `Wires` object with `device_wires.labels`. Is this correct?

mariaschuld: Yes, perfect! Well, strictly speaking the map goes from ``self.wires -> Wires(['q1', 'ancilla', 0, 1]) to device_wires -> Wires([0, 1, 2, 3])``, and ``device_wires.labels`` is the tuple ``(0, 1, 2, 3)``...

PR #40, opened 2020-08-15
Adding PasqalDevice

co9olguy:


pennylane_cirq/cirq_device.py
```diff
@@ -69,23 +70,30 @@ class CirqDevice(QubitDevice, abc.ABC):
...
-            self.qubits = qubits
         else:
-            self.qubits = [cirq.LineQubit(wire) for wire in device_wires.labels]
+            qubits = [cirq.LineQubit(idx) for idx in range(num_wires)]
+
+        # cirq orders the subsystems based on a total order defined on qubits.
+        # For consistency, this plugin uses that same total order
+        self.qubits = sorted(qubits)
+
+        super().__init__(wires, shots, analytic)
```
co9olguy, line 93: I had to move down the superclass initialization here, since it is the thing that creates the `wire_map`, but the `wire_map` can't be determined until the cirq qubits have been assigned

pennylane_cirq/cirq_device.py
```diff
@@ -149,7 +157,10 @@ def reset(self):
         # pylint: disable=missing-function-docstring
         super().reset()
 
-        self.circuit = cirq.Circuit()
+        if self.cirq_device:
+            self.circuit = cirq.Circuit(self.cirq_device)
```
co9olguy, line 161: Needed in order to use a non-default Cirq device (`PasqalVirtualDevice`)

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pennylane>=0.9
-cirq>=0.7
+google-api-core[grpc]<=1.14.0
+git+https://github.com/quantumlib/Cirq.git@master
```
co9olguy, line 3: Pasqal functionality now merged into Cirq master, but not released

tests/test_cirq_device.py
```diff
@@ -63,10 +63,21 @@ def test_outer_init_of_qubits(self):
 
         dev = CirqDevice(4, 100, False, qubits=qubits)
         assert len(dev.qubits) == 4
-        assert dev.qubits[0] == cirq.GridQubit(0, 0)
-        assert dev.qubits[1] == cirq.GridQubit(0, 1)
-        assert dev.qubits[2] == cirq.GridQubit(1, 0)
-        assert dev.qubits[3] == cirq.GridQubit(1, 1)
+        assert dev.qubits == qubits
+
+    def test_outer_init_of_qubits_unordered(self):
```
co9olguy, line 68: This was never being tested properly before

pennylane_cirq/cirq_device.py
```diff
@@ -226,3 +244,10 @@ def apply(self, operations, **kwargs):
         # Diagonalize the given observables
         for operation in rotations:
             self._apply_operation(operation)
+
+    def define_wire_map(self, wires):
```
co9olguy, line 248: Trying to get `ThreeDQubit`s to work uncovered a lot of uncovered edge cases at the interaction point between PennyLane's wires and Cirq's qubits

pennylane_cirq/cirq_device.py
```diff
@@ -51,9 +53,14 @@ class CirqDevice(QubitDevice, abc.ABC):
...
-            as wires. The wire number corresponds to the index in the list.
-            By default, an array of ``cirq.LineQubit`` instances is created.
+        qubits (List[cirq.Qubit]): A list of Cirq qubits that are used
+            as wires. By default, an array of ``cirq.LineQubit`` instances is created.
+            Wires are mapped to qubits using Cirq's internal mechanism for ordering
+            qubits. For example, if ``wires=2`` and ``qubits=[q1, q2]``, with
+            ``q1>q2``, then the wire indices 0 and 1 are mapped to q2 and q1, respectively.
+            If the user provides their own wire labels, e.g., ``wires=["alice", "bob"]``, and the
+            qubits are the same as the previous example, then "alice" would map to qubit q2
+            and "bob" would map to qubit q1.
```
co9olguy, line 63: Note: this has a big possibility to confuse users, but I'm not sure there's any better way to deal with it

pennylane_cirq/cirq_device.py
```diff
@@ -51,9 +53,14 @@ class CirqDevice(QubitDevice, abc.ABC):
...
     """
 
     name = "Cirq Abstract PennyLane plugin baseclass"
```
co9olguy, line 66:
```suggestion
    name = "Cirq Abstract PennyLane plugin base class"
```

pennylane_cirq/simulator_device.py
```diff
@@ -132,9 +132,10 @@ def apply(self, operations, **kwargs):
         # pylint: disable=missing-function-docstring
         super().apply(operations, **kwargs)
 
-        # We apply an identity gate to all wires, otherwise Cirq would ignore
+        # We apply identity gates to all wires, otherwise Cirq would ignore
         # wires that are not acted upon
-        self.circuit.append(cirq.IdentityGate(len(self.qubits))(*self.qubits))
```
co9olguy, line 137: This line broke the PasqalDevice, since it has restrictions about how close qubits must be in order to have multi-qubit gates. Note that it would break any other cirq device in the future which also had topology restrictions

pennylane_cirq/cirq_device.py
```diff
@@ -32,11 +32,13 @@
 ~~~~~~~~~~~~
 """
 import abc
+from collections import Iterable, OrderedDict
```
josh146, line 35:
```suggestion
from collections.abc import Iterable, OrderedDict
```

pennylane_cirq/cirq_device.py
josh146: This is super confusing 🤔

pennylane_cirq/cirq_device.py
josh146: @mariaschuld might have some ideas on how to improve `Wire`/cirq qubit integration (there was an allusion to it in #39 and #37)

pennylane_cirq/simulator_device.py
josh146: I'm wondering if applying identity gates to unused qubits is the right approach. It sounds like Cirq is being somewhat smart here, and simply just discarding qubits from the simulation that are not involved in the computation. Pyquil does this same; we should be able to simply take into account these ignored qubits when returning the samples/probabilities.

co9olguy: Yes, but I think this requires a change to how the devices work, doesn't it?

pennylane_cirq/cirq_device.py
co9olguy: spent most of a day trying to figure this out, the alternative is even more confusing :(

pennylane_cirq/simulator_device.py
josh146: I don't think so, the Forest simulators also act like this, and we take it into account in the PL plugin. We simply keep track of the inactive wires, and 'expand_dim' the probability before returning it.

co9olguy: Yes, then it makes sense to do that here. I think the previous solution (identity on all wires) was a quick hack.

But it might also make sense to provide helper functions for keeping track of such things in the base Device class?

pennylane_cirq/cirq_device.py
mariaschuld: I now I understand the following PR cherry picked from this one.

co9olguy:
@josh146 Not sure why CodeFactor is complaining about `Iterable`, it imports fine when I run locally :thinking:

josh146:
> @josh146 Not sure why CodeFactor is complaining about Iterable, it imports fine when I run locally 🤔

This is a known bug that is safe to ignore for now. The issue stems from the fact that `collections.Iterable` is deprecated, and the new home is `collections.abc.Iterable` -- this will be breaking in Python 3.9. Until then, the way Python makes this work from both import locations confuses pylint.

co9olguy:
docker build for `black` check erroring as well, but I think it's an external issue :thinking:

pennylane_cirq/cirq_device.py
```diff
@@ -32,9 +32,8 @@
 ~~~~~~~~~~~~
 """
 import abc
-from collections.abc import Iterable  # pylint: disable=no-name-in-module
+from collections.abc import Iterable
```
josh146, line 35:
```suggestion
from collections.abc import Iterable  # pylint: disable=no-name-in-module
```

requirements.txt
```diff
@@ -1,3 +1,3 @@
 pennylane>=0.11
-cirq>=0.7
+cirq>=0.9
```
josh146, line 2: We should update `setup.py` as well

josh146 (commented):
Looks good @co9olguy! A couple of high level comments before merging:

* the permissions changes on the `doc/xanadu_theme` folder should be reverted (`git checkout master doc/xanadu_theme`).

* The Pasqal device hasn't been added to the docs --- a card should be added on the `doc/index.rst` page, and a new page added in the `doc/devices/` directory. The information in the class docstring is not very user accessible (they would need to dig through the API to find it).

* We should update the setup.py requirements.

josh146:
> docker build for black check erroring as well, but I think it's an external issue 🤔

This is now fixed, it's failing still though because Black is complaining on an unformatted file :)

co9olguy:
> > docker build for black check erroring as well, but I think it's an external issue thinking
> 
> This is now fixed, it's failing still though because Black is complaining on an unformatted file :)

That is easy enough to deal with :)

doc/devices/pasqal.rst
```diff
@@ -0,0 +1,35 @@
...
+
+    @qml.qnode(dev)
+    def circuit(x):
+        qml.RX(x, wires=[0])
+        qml.CNOT(wires=[0, 1])
+        return qml.expval(qml.PauliZ(wires=1))
+
+    circuit(0.3)
+
+For more details about Pasqal devices, consult the `Cirq docs <https://cirq.readthedocs.io/en/stable/docs/pasqal/getting_started.html>`_.
```
josh146, line 35:
```suggestion
    circuit(0.3)

Note that if not specified, ``ThreeDGridQubits`` are automatically arranged in a linear
arrangement along the first coordinate axis, separated by a distance of ``control_radius / 2``.
That is, ``(0,0,0), (control_radius/2,0,0), (control_radius,0,0)``.

For more details about Pasqal devices, consult the `Cirq docs <https://cirq.readthedocs.io/en/stable/docs/pasqal/getting_started.html>`_.
```

josh146 (approved):
💯

PR #42, opened 2020-08-18
Fixes for device wire handling

josh146:
This PR contains the pasqal-independent bug fixes from #40 by @co9olguy

### DIFF
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -32,7 +32,7 @@ Code details
 ~~~~~~~~~~~~
 """
 import abc
-from collections import Iterable, OrderedDict
+from collections.abc import Iterable, OrderedDict
 
 import cirq
 import numpy as np
@@ -246,7 +246,7 @@ class CirqDevice(QubitDevice, abc.ABC):
         for operation in rotations:
             self._apply_operation(operation)
 
-    def define_wire_map(self, wires):
+    def define_wire_map(self, wires):  # pylint: disable=missing-function-docstring
         cirq_order = np.argsort(self._unsorted_qubits)
         consecutive_wires = Wires(cirq_order)
 
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -32,7 +32,8 @@ Code details
 ~~~~~~~~~~~~
 """
 import abc
-from collections.abc import Iterable, OrderedDict
+from collections.abc import Iterable  # pylint: disable=no-name-in-module
+from collections import OrderedDict
 
 import cirq
 import numpy as np
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -32,11 +32,14 @@ Code details
 ~~~~~~~~~~~~
 """
 import abc
+from collections import Iterable, OrderedDict
+
 import cirq
 import numpy as np
 import pennylane as qml
 from pennylane import QubitDevice
 from pennylane.operation import Operation
+from pennylane.wires import Wires
 
 from ._version import __version__
 from .cirq_operation import CirqOperation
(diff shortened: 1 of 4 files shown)


pennylane_cirq/cirq_device.py
```diff
@@ -69,23 +78,31 @@ class CirqDevice(QubitDevice, abc.ABC):
...
     def __init__(self, wires, shots, analytic, qubits=None):
-        super().__init__(wires, shots, analytic)
-
-        self.circuit = None
 
-        device_wires = self.map_wires(self.wires)
+        if not isinstance(wires, Iterable):
+            # interpret wires as the number of consecutive wires
+            wires = range(wires)
+        num_wires = len(wires)
```
mariaschuld, line 85: This should be done in ``super`` already. No harm in reproducing, but ``self.num_wires`` and ``self.wires`` should be defined at that stage! 

```suggestion
```

pennylane_cirq/cirq_device.py
```diff
@@ -69,23 +78,31 @@ class CirqDevice(QubitDevice, abc.ABC):
...
 
         if qubits:
-            if wires != len(qubits):
+            if num_wires != len(qubits):
```
mariaschuld, line 88:
```suggestion
            if len(self.wires) != len(qubits):
```

pennylane_cirq/cirq_device.py
```diff
@@ -69,23 +78,31 @@ class CirqDevice(QubitDevice, abc.ABC):
...
         else:
-            self.qubits = [cirq.LineQubit(wire) for wire in device_wires.labels]
+            qubits = [cirq.LineQubit(idx) for idx in range(num_wires)]
+
+        # cirq orders the subsystems based on a total order defined on qubits.
+        # For consistency, this plugin uses that same total order
+        self._unsorted_qubits = qubits
+        self.qubits = sorted(qubits)
+
+        super().__init__(wires, shots, analytic)
```
mariaschuld, line 102: I don't understand why `super` is not called earlier? That would remove a lot of code above, because `num_wires` can be extracted from `len(self.wires)` or `self.num_wires`. One could still raise an error if ``qubits`` has a different length.

pennylane_cirq/cirq_device.py
```diff
@@ -226,3 +246,10 @@ def apply(self, operations, **kwargs):
         # Diagonalize the given observables
         for operation in rotations:
             self._apply_operation(operation)
+
+    def define_wire_map(self, wires):  # pylint: disable=missing-function-docstring
+        cirq_order = np.argsort(self._unsorted_qubits)
+        consecutive_wires = Wires(cirq_order)
+
+        wire_map = zip(wires, consecutive_wires)
+        return OrderedDict(wire_map)
```
mariaschuld, line 255: Cool!

pennylane_cirq/simulator_device.py
```diff
@@ -132,9 +132,10 @@ def apply(self, operations, **kwargs):
         # pylint: disable=missing-function-docstring
         super().apply(operations, **kwargs)
 
-        # We apply an identity gate to all wires, otherwise Cirq would ignore
+        # We apply identity gates to all wires, otherwise Cirq would ignore
         # wires that are not acted upon
-        self.circuit.append(cirq.IdentityGate(len(self.qubits))(*self.qubits))
+        for q in self.qubits:
+            self.circuit.append(cirq.IdentityGate(1)(q))
```
mariaschuld, line 138: Not sure I understand why ``len(self.qubits)`` can be replaced by 1?

pennylane_cirq/cirq_device.py
josh146: @mariaschuld, here is @co9olguy's original comment: https://github.com/PennyLaneAI/pennylane-cirq/pull/40#discussion_r471025961

> I had to move down the superclass initialization here, since it is the thing that creates the wire_map, but the wire_map can't be determined until the cirq qubits have been assigned

mariaschuld: Ah maybe that is due to other changes made in the original PR? In that case we can merge as is, to not to create conflicts!

josh146:
Thanks @mariaschuld! I'll get @co9olguy to give it a once over before we merge it in.

pennylane_cirq/simulator_device.py
josh146: In this case, it looks like you are applying `len(self.qubit)` single-qubit identity gates, instead of one large identity gate.

pennylane_cirq/cirq_device.py
co9olguy: `define_wire_map` is called in the parent class's `__init__` method. 

We don't need to know the number of qubits (as you say, that can be inferred), but we do need to do all this qubit pre-processing before `define_wire_map` makes sense to be called

pennylane_cirq/cirq_device.py
co9olguy: Worked really nicely once I figured out that Cirq always uses an ordering internally (the edge case of unordered qubits was never caught before in this plugin)

pennylane_cirq/simulator_device.py
co9olguy: Yes, but this is just putting a band-aid over an existing hack. Would be better for us to keep better track of the unused wires in the plugin (which we can now do easily!)

pennylane_cirq/simulator_device.py
```diff
@@ -132,9 +132,10 @@ def apply(self, operations, **kwargs):
         # pylint: disable=missing-function-docstring
         super().apply(operations, **kwargs)
 
-        # We apply an identity gate to all wires, otherwise Cirq would ignore
+        # We apply identity gates to all wires, otherwise Cirq would ignore
```
co9olguy, line 135:
```suggestion
        # TODO: remove the need for this hack by keeping better track of unused wires
        # We apply identity gates to all wires, otherwise Cirq would ignore
```

pennylane_cirq/simulator_device.py
co9olguy: Have added a TODO in the codebase

PR #45, opened 2020-10-30
Add iSWAP and CPhase gates

thisac:
This PR adds the iSWAP and CPhase gates from Cirq to `ops.py`.


josh146 (approved):
Looks good from my end Theo :+1:

Do we have tests for the plugin custom ops?

pennylane_cirq/ops.py
```diff
@@ -123,3 +123,43 @@ class Depolarize(Operation):
...
+    Implemented as Cirq ``CZPowGate(exponent=phi / np.pi)``.
+
+    .. math::
+
+        CPhase(\phi) =
+            \begin{bmatrix}
+                [[1, 0, 0, 0],
+                [0, 1, 0, 0],
+                [0, 0, 1, 0],
+                [0, 0, 0, e^{i\phi]]
```
co9olguy, line 154:
```suggestion
                [0, 0, 0, e^{i\phi}]]
```

co9olguy: Doesn't the `bmatrix` need newlines?

tests/test_native_ops.py
```diff
@@ -188,3 +188,54 @@ def test_apply_amplitude_damp_single_wire(
...
+    @pytest.mark.parametrize("input",
+        [
+            [1, 0, 0, 0],
+            [0, 1, 0, 0],
+            [0, 0, 1, 0],
+            [0, 0, 0, 1],
+        ]
+    )
+    def test_apply_iswap(self, tol, input, shots, analytic):
+        """Tests that applying the CPhase gate yields the expected output."""
```
co9olguy, line 201: In both of these new tests, the docstring and test name do not match

tests/test_native_ops.py
```diff
@@ -188,3 +188,54 @@ def test_apply_amplitude_damp_single_wire(
...
+        device = SimulatorDevice(2, shots=shots, analytic=analytic)
+
+        iswap_mat = np.array([[1, 0, 0, 0],
+                              [0, 0, 1j, 0],
+                              [0, 1j, 0, 0],
+                              [0, 0, 0, 1]])
+
+        expected = input @ iswap_mat
```
co9olguy, line 209: Why is the gate being applied to the RHS? Is this only working because the input is the identity matrix (and hence commutes with all gates)?

tests/test_native_ops.py
```diff
@@ -188,3 +188,54 @@ def test_apply_amplitude_damp_single_wire(
...
         assert np.allclose(simulator_device_1_wire.state, expected_density_matrix, **tol)
+
+    @pytest.mark.parametrize("input",
+        [
+            [1, 0, 0, 0],
+            [0, 1, 0, 0],
+            [0, 0, 1, 0],
+            [0, 0, 0, 1],
+        ]
+    )
```
co9olguy, line 199: I struggled to understand this. If the device has two wires, then shouldn't the underlying pure state have shape `(2,2)`?

co9olguy (changes requested):
Thanks @thisac. I found a few obvious typos and have some questions about the tests.

tests/test_native_ops.py
thisac: It should have shape `(4,)`, right? For the different bitstrings `00`, `01`, `10` and `11`. Which is what this is (each row corresponding to an input). :thinking:

tests/test_native_ops.py
thisac: :facepalm:

tests/test_native_ops.py
thisac: :facepalm: I shouldn't have attempted fixing this so late yesterday night.

pennylane_cirq/ops.py
thisac: You're completely right. This is a very ugly mix of Python and Latex. :laughing:

thisac:
Thanks @co9olguy for the review. Too many :facepalm:s in this one though, but I fixed the typos and the gate application order, and also tried to make the parametrization less confusing (adding a few more interesting states, and avoiding it looking like the identity).

tests/test_native_ops.py
co9olguy: Aha, got it! the pytest parametrize format looks exactly like a 2d array :laughing:

co9olguy: Might be good to test on superposition states as well

tests/test_native_ops.py
```diff
@@ -191,22 +191,24 @@ def test_apply_amplitude_damp_single_wire(
...
-            [1, 0, 0, 0],
-            [0, 1, 0, 0],
-            [0, 0, 1, 0],
-            [0, 0, 0, 1],
+            np.array([1, 0, 0, 0]),
+            np.array([2, 1, 0, 1]) / np.sqrt(6),
+            np.array([0, 0, 1, 0]),
+            np.array([0, 1, 0, 1]) / np.sqrt(2),
+            np.array([0, 0, 0, 1]),
+            np.array([2, 1, 2, 1]) / np.sqrt(10),
```
co9olguy, line 199: :100:

co9olguy (approved):

PR #46, opened 2020-11-09
Remove import of qsimcirq from `__init__`

thisac:
Removes the import of `qsimcirq` from the `__init__.py` file and fixes the tests and entrypoints to import qsimcirq correctly.

This avoids users having to install `qsimcirq` for the other Cirq devices to work.

### DIFF
--- a/pennylane_cirq/__init__.py
+++ b/pennylane_cirq/__init__.py
@@ -16,7 +16,6 @@ Plugin overview
 ===============
 """
 from .simulator_device import SimulatorDevice, MixedStateSimulatorDevice
-from .qsim_device import QSimDevice, QSimhDevice
 from .pasqal_device import PasqalDevice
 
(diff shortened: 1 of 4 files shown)


co9olguy (approved):

josh146:
@thisac don't forget to also update the changelog! As the number of plugins increases, this becomes more important so as to ensure that we can keep track of changes + when pypi releases are required

co9olguy:
@josh146 I wonder if we can set up a GitHub action that checks whether there are any updates to CHANGELOG?

josh146:
Yes! @mariaschuld and I were discussing this recently.

I think the best approach is to have a bot comment a reminder when PRs are opened, rather than have a check that must pass for merging (avoids complexity, and allows for trivial PRs where we don't want to necessarily update the changelog).

See https://github.com/marketplace/actions/changelog-reminder:

image

PR #51, opened 2021-02-18
Added custom simulator option

Thenerdstation:
Allows users to use custom cirq simulators with the existing cirq plugin.

Main reason for this is to support the new Floq integration. To use floq, users simply now can do

```python
import pennylane as qml
import remote_cirq 

floq_sim = remote_cirq.RemoteSimulator("MY_API_KEY")

dev = qml.device("cirq.simulator", wires=2, simulator=floq_sim)

@qml.qnode(dev)
def my_circuit():
    # ....
    return qml.expval(...)
```

And that's it!


pennylane_cirq/cirq_device.py
```diff
@@ -155,14 +158,29 @@ def __init__(self, wires, shots, analytic, qubits=None):
...
-        "PauliZ": None,
-        "Hadamard": None,
+        "PauliX": CirqOperation(lambda: cirq.X),
+        "PauliY": CirqOperation(lambda: cirq.Y),
+        "PauliZ": CirqOperation(lambda: cirq.Z),
+        "Hadamard": CirqOperation(lambda: cirq.H),
         "Hermitian": None,
-        "Identity": None,
+        "Identity": CirqOperation(lambda: cirq.I),
     }
```
josh146, line 167: This dictionary is actually useful now

pennylane_cirq/cirq_device.py
```diff
@@ -155,14 +158,29 @@ def __init__(self, wires, shots, analytic, qubits=None):
...
 
+    def to_paulistring(self, observable):
+        """Convert an observable to a cirq.PauliString"""
+        if isinstance(observable, Tensor):
+            obs = [self.to_paulistring(o) for o in observable.obs]
+            return functools.reduce(operator.mul, obs)
+
+        cirq_op = self._observable_map[observable.name]
+        if cirq_op is None:
+            raise NotImplementedError(f"{observable.name} is not currently supported.")
```
josh146, line 177: I assume this catches `qml.Hermitian`?

Out of curiosity, are there any observables cirq supports that PL doesn't?

josh146: We _could_ use `qml.utils.decompose_hamiltonian()` to convert the Hermitian matrix to a Pauli string, but not sure if it's worth it

pennylane_cirq/simulator_device.py
```diff
@@ -188,6 +188,14 @@ def generate_samples(self):
             [self._result.measurements[str(wire)].flatten() for wire in range(self.num_wires)]
         ).T.astype(int)
 
+    def expval(self, observable):
+        # pylint: disable=missing-function-docstring
+        if not self.analytic or not hasattr(self._simulator, "simulate_expectation_values"):
+            return super().expval(observable)
+        return self._simulator.simulate_expectation_values(
+            self.circuit, self.to_paulistring(observable)
+        )[0]
```
josh146, line 197: Nice, I suppose this leads to a performance improvement?

tests/test_simulator_device.py
```diff
@@ -37,6 +37,20 @@ def test_device_loading(self):
...
+        sim = cirq.Simulator()
+        sim.simulate_expectation_values = MagicMock(return_value=[0.123])
+        dev = qml.device("cirq.simulator", wires=1, simulator=sim)
+
+        @qml.qnode(dev)
+        def circuit():
+            qml.PauliX(0)
+            return qml.expval(qml.PauliX(0))
+        
+        assert circuit() == 0.123
```
josh146, line 51: nice!

josh146 (approved):
Looks good from my end! Don't forget to add a changelog entry, just so we can keep track for the next release

pennylane_cirq/simulator_device.py
```diff
@@ -59,11 +59,11 @@ class SimulatorDevice(CirqDevice):
     """
     name = "Cirq Simulator device for PennyLane"
     short_name = "cirq.simulator"
-
-    def __init__(self, wires, shots=1000, analytic=True, qubits=None):
+    # pylint: disable=too-many-arguments
+    def __init__(self, wires, shots=1000, analytic=True, qubits=None, simulator=None):
```
josh146, line 63: Oh, also don't forget to add the new keyword argument to the docstring

pennylane_cirq/cirq_device.py
Thenerdstation: lol

pennylane_cirq/simulator_device.py
Thenerdstation: I would but there wasn't a docstring before. I can add one now tho.

pennylane_cirq/simulator_device.py
Thenerdstation: Yes, it essentially moves the responsibility of calculating the expectation value to the cirq device rather than just pennylane. The downstream device can then optimize how it's calculated.

pennylane_cirq/cirq_device.py
Thenerdstation: I can add a TODO

pennylane_cirq/simulator_device.py
josh146: There's one above, below the `class :` line :)

pennylane_cirq/simulator_device.py
josh146: Awesome. A lot of PennyLane plugin patterns are repeated from a time (a long way!) back when other frameworks didn't provide (accessible) expectation APIs. Now that this is commonplace, we should take this into account more when writing plugins.

pennylane_cirq/simulator_device.py
Thenerdstation: oh i see

pennylane_cirq/simulator_device.py
Thenerdstation: Agreed, especially for more future remote services (both simulation and real QC).

PR #55, opened 2021-02-22
Replace cirq.rx etc by XPowGate etc.

dwierichs:
This will enable floq to use qml.RX, qml.RY, qml.RZ, qml.CRX, qml.CRZ via the PowGates instead of the corresponding cirq implementations or RX etc, which do not work with floq.
CRY and CRot will not work unless one explicitly decomposes them or implements CYPowGate (for both).
In order to make the gates work, I commented out the corresponding operations in the device such that they should be decomposed by pennylane.

### DIFF
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -142,7 +142,6 @@ class CirqDevice(QubitDevice, abc.ABC):
         "RX": CirqOperation(lambda phi: cirq.XPowGate(exponent=phi / np.pi, global_shift=-0.5)),
         "RY": CirqOperation(lambda phi: cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5)),
         "RZ": CirqOperation(lambda phi: cirq.ZPowGate(exponent=phi / np.pi, global_shift=-0.5)),
-cirq.ZPowGate(exponent=a / np.pi, global_shift=-0.5)
         "Rot": CirqOperation(lambda a, b, c: [cirq.ZPowGate(exponent=a / np.pi, global_shift=-0.5), cirq.YPowGate(exponent=b / np.pi, global_shift=-0.5), cirq.ZPowGate(exponent=c / np.pi, global_shift=-0.5)]),
         "CRX": CirqOperation(lambda phi: cirq.ControlledGate(cirq.XPowGate(exponent=phi / np.pi, global_shift=-0.5))),
         "CRY": CirqOperation(lambda phi: cirq.ControlledGate(cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5))),
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -143,9 +143,9 @@ class CirqDevice(QubitDevice, abc.ABC):
         "RY": CirqOperation(lambda phi: cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5)),
         "RZ": CirqOperation(lambda phi: cirq.ZPowGate(exponent=phi / np.pi, global_shift=-0.5)),
         "Rot": CirqOperation(lambda a, b, c: [cirq.ZPowGate(exponent=a / np.pi, global_shift=-0.5), cirq.YPowGate(exponent=b / np.pi, global_shift=-0.5), cirq.ZPowGate(exponent=c / np.pi, global_shift=-0.5)]),
-        "CRX": CirqOperation(lambda phi: cirq.ControlledGate(cirq.XPowGate(exponent=phi / np.pi, global_shift=-0.5))),
-        "CRY": CirqOperation(lambda phi: cirq.ControlledGate(cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5))),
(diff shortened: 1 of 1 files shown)


pennylane_cirq/cirq_device.py
```diff
@@ -139,20 +139,26 @@ def __init__(self, wires, shots, analytic, qubits=None):
...
+        "CRX": CirqOperation(lambda phi: cirq.CNotPowGate(exponent=phi / np.pi, global_shift=-0.5)),
+        # "CRY": CirqOperation(lambda phi: cirq.ControlledGate(cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5))),
+        "CRZ": CirqOperation(lambda phi: cirq.CZPowGate(exponent=phi / np.pi, global_shift=-0.5)),
+        # "CRot": CirqOperation(
+        # lambda a, b, c: [
+        # cirq.ControlledGate(cirq.rz(a)),
+        # cirq.ControlledGate(cirq.ry(b)),
+        # cirq.ControlledGate(cirq.rz(c)),
+        # ]
+        # ),
```
Thenerdstation: Let's keep these operations enabled for the non-floq usage.

pennylane_cirq/cirq_device.py
```diff
@@ -139,20 +139,26 @@ def __init__(self, wires, shots, analytic, qubits=None):
...
-        "RY": CirqOperation(cirq.ry),
-        "RZ": CirqOperation(cirq.rz),
-        "Rot": CirqOperation(lambda a, b, c: [cirq.rz(a), cirq.ry(b), cirq.rz(c)]),
-        "CRX": CirqOperation(lambda phi: cirq.ControlledGate(cirq.rx(phi))),
-        "CRY": CirqOperation(lambda phi: cirq.ControlledGate(cirq.ry(phi))),
-        "CRZ": CirqOperation(lambda phi: cirq.ControlledGate(cirq.rz(phi))),
-        "CRot": CirqOperation(
+        "RX": CirqOperation(lambda phi: cirq.XPowGate(exponent=phi / np.pi, global_shift=-0.5)),
+        "RY": CirqOperation(lambda phi: cirq.YPowGate(exponent=phi / np.pi, global_shift=-0.5)),
+        "RZ": CirqOperation(lambda phi: cirq.ZPowGate(exponent=phi / np.pi, global_shift=-0.5)),
```
Thenerdstation: Do these need to be replaced? They worked fine when I tested them.

dwierichs: I thought they were faulty, but maybe that was miscommunication on my end, sorry!
I am kind of confused by the global phase factor...

pennylane_cirq/cirq_device.py
dwierichs: Okay. I thought when using them with floq, they'd then just be decomposed. Optimally, one would just work on the floq device, I guess?

pennylane_cirq/cirq_device.py
```diff
@@ -139,20 +139,26 @@ def __init__(self, wires, shots, analytic, qubits=None):
...
+        "RZ": CirqOperation(lambda phi: cirq.ZPowGate(exponent=phi / np.pi, global_shift=0.5)),
+        "Rot": CirqOperation(
             lambda a, b, c: [
-                cirq.ControlledGate(cirq.rz(a)),
-                cirq.ControlledGate(cirq.ry(b)),
-                cirq.ControlledGate(cirq.rz(c)),
+                cirq.ZPowGate(exponent=a / np.pi, global_shift=0.5),
+                cirq.YPowGate(exponent=b / np.pi, global_shift=0.5),
+                cirq.ZPowGate(exponent=c / np.pi, global_shift=0.5),
             ]
```
Thenerdstation, line 150: I'm not sure we can replace these 2 qubit gates with a 1 qubit gates

Thenerdstation:
Make sure the tests are all passing too. Most of the failures are numeric issues, but it could be possible that they are due to global phase.

pennylane_cirq/cirq_device.py
dwierichs: Rot is a 1 qubit gate, right? I don't understand the diff..

dwierichs:
Yeah, sorry, I'm working on it. I used the cirq comment on how to convert between rotation  and pow gates, but this seems to be off, the probabilities are fine but the phases are not. I'll put it on WIP

Thenerdstation:
Phases might me less of an issue. RX and XPowGate are just off by a constant global phase, so it might not be _wrong_ but we'd need to update the tests to correct it.

dwierichs:
Yeah, I agree. But the cirq gates seem to allow for correcting this, so I'm trying to use that feature to not have to modify the tests :)

dwierichs:
Okay, so the difference is not a global phase for the controlled rotation, but a controlled phase, which amounts to PhaseShift(-theta/2, wires=wires[0]) when wires[0] is the control qubit.
In order to implement this, the pennylane-cirq.CirqOperation would need to allow for an iterable to contain gates that act on different numbers of wires, which I tried to implement but I failed at that.
I have no idea how to easily circumvent this. The doc in cirq should definitely be updated.

Jaybsoni:
Closing due to inactivity

PR #56, opened 2021-03-08
Remove custom expval path

Thenerdstation:
Fixes broken build

### DIFF
--- a/pennylane_cirq/simulator_device.py
+++ b/pennylane_cirq/simulator_device.py
@@ -190,14 +190,6 @@ class SimulatorDevice(CirqDevice):
             [self._result.measurements[str(wire)].flatten() for wire in range(self.num_wires)]
         ).T.astype(int)
 
-    def expval(self, observable):
-        # pylint: disable=missing-function-docstring
-        if not hasattr(self._simulator, "simulate_expectation_values"):
-            return super().expval(observable)
-        return self._simulator.simulate_expectation_values(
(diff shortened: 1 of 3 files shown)


pennylane_cirq/simulator_device.py
```diff
@@ -190,14 +190,6 @@ def generate_samples(self):
             [self._result.measurements[str(wire)].flatten() for wire in range(self.num_wires)]
         ).T.astype(int)
 
-    def expval(self, observable):
-        # pylint: disable=missing-function-docstring
-        if not hasattr(self._simulator, "simulate_expectation_values"):
-            return super().expval(observable)
-        return self._simulator.simulate_expectation_values(
-            self.circuit, cirq.PauliSum() + self.to_paulistring(observable)
-        )[0]
```
josh146, line 199: Would we want to keep this, but simply support the new `simulate_expectation_values` API?

It would be easier to support this though onces Shots II has been merged

Thenerdstation: Agreed. I think I'll add that in post shots 2 refactor. This code was only for Qhack/Floq (and based on the user feedback, I was wrong in ignoring the shots value here).

josh146 (approved):
Thanks @Thenerdstation! This is nice to fix ASAP; I imagine we might get some bug reports in, and we can direct those users to install from github via
```python
pip install git+https://github.com/PennyLaneAI/pennylane-cirq
```
temporarily

Issue #53, opened 2021-02-22
Update expectation value computation logic 


josh146:
Currently, the simulator device uses the following logic for expectation values:

```python
def expval(self, observable):
    if not hasattr(self._simulator, "simulate_expectation_values"):
        return super().expval(observable)
     return self._simulator.simulate_expectation_values(
        self.circuit, cirq.PauliSum() + self.to_paulistring(observable)
     )[0]
```

That is, if the simulator has the `simulate_expectation_values` attribute, this is used to compute the **exact** expectation values.

Note that **this ignores the shots/analytic argument**. So if the attribute exists, exact expectation values are always returned.

Pre-v0.10 of Cirq, this is fine, since `cirq.Simulator` does not have this attribute. However, we will need to revisit as soon as:

* The shots refactor is complete
* Cirq v0.10 is out.

rmoyard:
This issue is solved by #81 .

PR #57, opened 2021-03-24
Analytic fix

Thenerdstation:
Fixes to account for the removal of the `analytic` keyword.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -362,8 +362,8 @@ inheritance_node_attrs = dict(color='lightskyblue1', style='filled')
 #autodoc_default_flags = ['members']
 autosummary_generate = True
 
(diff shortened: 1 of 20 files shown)


josh146 (approved):
Thanks @Thenerdstation 💯 

Does this close #53, or is that still open to some degree? iirc, maybe that is no longer an issue, since we returned to the manual approach of computing expvals for samples always, rather than asking Cirq for the expval directly.

Perhaps now we can start to add that back in.

Thenerdstation:
I would consider it still open, as this PR doesn't add the `simulate_expectation_values` method from Cirq.

Thenerdstation:
Should I merge with the git link in the requirements?

josh146:
yep go for it

PR #58, opened 2021-04-12
Update simulator_device.py

Thenerdstation:
shots=1000 -> shots=None


trbromley (changes requested):
Thanks @Thenerdstation! We also need this on the other devices:
- `QSimDevice`
- `QSimhDevice`
- `PasqalDevice`
- `MixedStateSimulatorDevice`

Also, what about adding a quick test? I was surprised the tests were still passing with non-analytic being the default, but this is because we never check the default for `shots` and always explicitly pass. So even a test where we don't pass a `shots` argument but we check that `dev.shots is None`?

PR #61, opened 2021-05-26
Fix issue when using a wire subset with `BasisState`

thisac:
Fixes an issue when calling `BasisState` when using a subset of the device wires.

The following now works:
```python
dev = qml.device("cirq.simulator", wires=2)

@qml.qnode(dev)
def circuit():
    qml.BasisState([1], wires=1)
    return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1))

circuit()
```
Previously it raised 
`DeviceError: For BasisState, the state has to be specified for the correct number of qubits. Got a state for 1 qubits, expected 2.` 

and now it should output the correct `[1, -1]`.

### DIFF
--- a/pennylane_cirq/simulator_device.py
+++ b/pennylane_cirq/simulator_device.py
@@ -49,10 +49,7 @@ class SimulatorDevice(CirqDevice):
             or strings (``['ancilla', 'q1', 'q2']``).
         shots (int): Number of circuit evaluations/random samples used
             to estimate expectation values of observables. Shots need
-            to >= 1. In analytic mode, shots indicates the number of entries
-            that are returned by ``device.sample``.
-        analytic (bool): Indicates that expectation values and variances should
-            be calculated analytically. Defaults to ``True``.
+            to >= 1. If ``None``, expectation values are calculated analytically.
         qubits (List[cirq.Qubit]): A list of Cirq qubits that are used
             as wires. The wire number corresponds to the index in the list.
             By default, an array of ``cirq.LineQubit`` instances is created.
@@ -62,8 +59,8 @@ class SimulatorDevice(CirqDevice):
     name = "Cirq Simulator device for PennyLane"
     short_name = "cirq.simulator"
     # pylint: disable=too-many-arguments
-    def __init__(self, wires, shots=1000, analytic=True, qubits=None, simulator=None):
-        super().__init__(wires, shots, analytic, qubits)
+    def __init__(self, wires, shots=1000, qubits=None, simulator=None):
+        super().__init__(wires, shots, qubits)
 
         self._simulator = simulator or cirq.Simulator()
 
@@ -81,7 +78,7 @@ class SimulatorDevice(CirqDevice):
 
     def _apply_basis_state(self, basis_state_operation):
         # pylint: disable=missing-function-docstring
-        if not self.analytic:
+        if not self.shots is None:
             raise qml.DeviceError("The operation BasisState is only supported in analytic mode.")
 
         basis_state_array = np.array(basis_state_operation.parameters[0])
@@ -106,7 +103,7 @@ class SimulatorDevice(CirqDevice):
 
     def _apply_qubit_state_vector(self, qubit_state_vector_operation):
         # pylint: disable=missing-function-docstring
-        if not self.analytic:
+        if not self.shots is None:
             raise qml.DeviceError(
                 "The operation QubitStateVector is only supported in analytic mode."
             )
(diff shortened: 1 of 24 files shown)


pennylane_cirq/simulator_device.py
```diff
@@ -109,11 +133,18 @@ def _apply_qubit_state_vector(self, qubit_state_vector_operation):
             )
 
         state_vector = np.array(qubit_state_vector_operation.parameters[0], dtype=np.complex64)
+        try:
+            wires = self.map_wires(qubit_state_vector_operation.wires)
+        except WireError:
+            wires = qubit_state_vector_operation.wires
```
thisac, line 139: I'm not sure why this is, but there are cases when `QubitStateVector` is called on more wires than the device has. E.g., where it's called on wires `[0, 1]` but the device only has 1 wire.

```python
def test_var_single_wire_no_parameters(
    self, simulator_device_1_wire, tol, operation, input, expected_output
):
    """Tests that variances are properly calculated for single-wire observables without parameters."""

    op = operation(0, do_queue=False)

    simulator_device_1_wire.reset()
    simulator_device_1_wire.apply(
        [qml.QubitStateVector(np.array(input), wires=[0, 1])],
        rotations=op.diagonalizing_gates(),
    )
```

This causes issues with `map_wires` (thus the try-except) and also the `_expand_state` below (thus the `if len(wires) < self.num_wires ...` instead of `if len(wires) != self.num_wires ...`, as I first wrote).

josh146: @thisac I would consider this a bug in the tests :) I suggest:

- removing the try-except here
- updating the test to use a 2 wire device

josh146 (commented):
I did a quick skim through, looks good so far @thisac! I haven't done a proper review yet, as I noticed there are no tests (yet) to ensure that this bug is fixed

pennylane_cirq/simulator_device.py
```diff
@@ -81,26 +83,47 @@ def _apply_basis_state(self, basis_state_operation):
...
                     basis_state_operation.parameters[0]
                 )
             )
 
+        # expand basis state to device wires
+        basis_state_array = np.zeros(self.num_wires, dtype=int)
+        basis_state_array[wires] = basis_state_operation.parameters[0]
+
         self._initial_state = np.zeros(2 ** len(self.qubits), dtype=np.complex64)
         basis_state_idx = np.sum(2 ** np.argwhere(np.flip(basis_state_array) == 1))
```
josh146, line 107: Oh, this is a permutation of converting from base2 to base10 that I hadn't seen before!

Interestingly, however, this seems to be a rare case where the native Python approach is still faster:

```pycon
>>> bs = np.array([1, 1, 0, 1, 1, 0, 0, 1, 1])
>>> int("".join(str(i) for i in bs), 2)
435
>>> np.sum(2 ** np.argwhere(np.flip(bs) == 1))
435
>>> %timeit np.sum(2 ** np.argwhere(np.flip(bs) == 1))
41.4 µs ± 4.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit int("".join(str(i) for i in bs), 2)
14.4 µs ± 489 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

pennylane_cirq/simulator_device.py
```diff
@@ -81,26 +83,47 @@ def _apply_basis_state(self, basis_state_operation):
...
+        unravelled_indices = np.zeros((2 ** len(wires), self.num_wires), dtype=int)
+        unravelled_indices[:, wires] = basis_states
+
+        # get indices for which the state is changed to input state vector elements
+        ravelled_indices = np.ravel_multi_index(unravelled_indices.T, [2] * self.num_wires)
+
+        state_vector = self._scatter(ravelled_indices, state_vector, [2 ** self.num_wires])
+        state_vector = self._reshape(state_vector, [2] * self.num_wires)
+        state_vector = self._asarray(state_vector, dtype=self.C_DTYPE)
+
```
josh146, line 124: Same comment here as over on the Qulacs PR, recommend changing this to simply use native NumPy functions :)

tests/test_simulator_device.py
```diff
@@ -485,6 +485,27 @@ def test_qubit_state_vector_not_at_beginning_error(self, simulator_device_1_wire
...
+        (np.array([1, 2, 3, 4]) / np.sqrt(48), 3, [1, 2], np.array([1, 2, 3, 4, 0, 0, 0, 0]) / np.sqrt(48)),
+    ],
+)
+@pytest.mark.parametrize("shots", [None])
+def test_expand_state(state, op_wires, device_wires, expected, tol):
+    """Test that the expand_state method works as expected."""
+    dev = SimulatorDevice(device_wires)
+    res = dev._expand_state(state, op_wires)
+
+    assert np.allclose(res, expected, **tol)
```
josh146, line 507: great tests!

josh146 (approved):
Thanks @thisac! Tests look very nice.

Only suggestion is to use standard NumPy rather than the device static methods.

PR #62, opened 2021-05-27
Add a projector observable to CirqDevice

wongwsvincent:
Context:
Adding a projector observable to CirqDevice.

Description of the Change:
Implementing projector observable with Cirq's own ```cirq.ProductState.projector``` method.

Related GitHub PR:
Work coherently with PR [#1356](https://github.com/PennyLaneAI/pennylane/pull/1356)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
 
     steps:
       - name: Cancel Previous Runs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## New features
 
-* Added Projector observable to CirqDevice.
+* Added support for the new `qml.Projector` observable in
+  PennyLane v0.16 to the Cirq devices.
   [(#62)](https://github.com/PennyLaneAI/pennylane-cirq/pull/62)
 
 ## Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## New features
 
+* Added Projector observable to CirqDevice.
+  [(#62)](https://github.com/PennyLaneAI/pennylane-cirq/pull/62)
+
 ## Improvements
 
 ## Bug fixes
@@ -13,6 +16,8 @@
 ### Contributors
 
 This release contains contributions from (in alphabetical order):
+
+Vincent Wong
 
 ---
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git
 cirq>=0.10
 numpy~=1.16
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -166,7 +166,7 @@ class CirqDevice(QubitDevice, abc.ABC):
         # to support this observable.
         "Hermitian": None,
         "Identity": CirqOperation(lambda: cirq.I),
-        "Projector": CirqOperation(lambda: cirq.ProductState.projector)
+        "Projector": CirqOperation(lambda: cirq.ProductState.projector),
     }
 
     def to_paulistring(self, observable):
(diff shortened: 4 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## New features
 
+* Added Projector observable to CirqDevice.
```
josh146, line 5:
```suggestion
* Added support for the new `qml.Projector` observable in
  PennyLane v0.16 to the Cirq devices.
```

(Just adding a little more details for readers)

josh146 (changes requested):
Thanks @wongwsvincent! Before I can approve, would you be able to add a small test for the projector observable, to make sure that it works correctly? Since it will be using `Cirq` directly to compute the expectation, rather than using the logic you added to `QubitDevice` (I believe).

I think there are two options here:

- Add a small unit test using the device methods directly
- Write an integration test using a QNode (this would require modifying the `requirements.txt` file so that the latest master of PennyLane is installed instead, e.g.

  ```python
  git+https://github.com/PennyLaneAI/pennylane.git
  ```

Thank you for adding this!

wongwsvincent:
@josh146,  I have only added a ```test_projector_expectation``` function in ```tests/test_expval.py``` as a start, as I have trouble setting up ```qsimcirq``` on my Mac, so I can't run ```make test``` locally and I'm hesitating a bit about whether I should continue committing tests I haven't checked myself locally.

P.S. It would be great if there's any suggestion on how to install ```qsimcirq``` on Mac OS, but I understand this is not an installation related to PennyLane library.

josh146:
Thanks @wongwsvincent!

> as I have trouble setting up qsimcirq on my Mac, so I can't run make test locally and I'm hesitating a bit about whether I should continue committing tests I haven't checked myself locally.

This is perfectly fine 🙂 If you want to run the test suite locally, you can explicitly ignore the `qsim` tests, either by only running the test file you have modified,

```
pytest tests/test_expval.py -x --tb=short
```

or by using pytest's [test selector command line argument](https://docs.pytest.org/en/6.2.x/usage.html#specifying-tests-selecting-tests):

```
pytest tests -x --tb=short -k "not (test_qsim or test_qsimh)"
```

> P.S. It would be great if there's any suggestion on how to install qsimcirq on Mac OS, but I understand this is not an installation related to PennyLane library.

Unfortunately I am not sure - I know that `qsim` is quite tricky to install, even on linux! `qsim` _might_ provide a dockerfile to use it with Docker, but apart from that, I am unsure. Note that it is perfectly fine to not run the test locally, and instead rely on CI within the PR :)

wongwsvincent:
Awesome! Thanks @josh146 , something like ```pytest tests/test_expval.py -x --tb=short``` is exactly what I need. 
I ran that command and it passed the tests successfully. The same tests as the ones I added in ```pennylane/devices/tests/test_measurements.py``` are now included in ```pennylane-cirq/tests/test_expval.py```, ```pennylane-cirq/tests/test_sample.py``` and ```pennylane-cirq/tests/test_var.py```.
Hopefully all the tests on CI will be fine, and the PR will be ready to merge.

.github/workflows/tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
```
josh146, line 13: I removed 3.6, since PL no longer supports it!

requirements.txt
```diff
@@ -1,3 +1,3 @@
-pennylane>=0.15
+git+https://github.com/PennyLaneAI/pennylane.git
```
josh146, line 1: I also updated this to point to PL master, since `qml.Projector` is not present in the stable v0.15 release

josh146 (approved):
Thanks @wongwsvincent for the extra push with the tests!

The integration tests are still failing, but this is not the fault of your PR, rather, it looks like another bug in the Cirq plugin has been uncovered. Since that is out of scope here, I may override that failing CI check.

wongwsvincent:
Thanks @josh146 for fixing some of the test errors! 
There are six remaining errors all complaining about the use of ```numpy.complex64``` vs ```complex128```, but they don't look like they are caused by this PR as you have pointed out. 
Please let me know if there's anything I am missing or I can help with this PR.

josh146:
> There are six remaining errors all complaining about the use of numpy.complex64 vs complex128, but they don't look like they are caused by this PR as you have pointed out. Please let me know if there's anything I am missing or I can help with this PR.

Yep, they are orthogonal :) I will merge this PR now @wongwsvincent, which allows us to fix these bugs in a separate PR. I have to admit, off the top of my head I'm unsure what the bug is, it will require some investigation.

PR #63, opened 2021-06-14
Fix bug for mixed simulator

rmoyard:
This data type change should fix the bug appearing when building the density matrix for `QubitStateVector`.


thisac (approved):
Looks great @rmoyard! 💯

PR #87, opened 2021-11-18
Fix test for `num_param` refractor

rmoyard:
The recent refractor of the `Operator` class made the use of `num_params` impossible before constructing the operation, therefore a test that builds random circuit has to be changed.

### DIFF
--- a/tests/test_mixed_simulator_device.py
+++ b/tests/test_mixed_simulator_device.py
@@ -708,26 +708,26 @@ class TestExpval:
         dev = qml.device("cirq.mixedsimulator", wires=4)
 
         gates = [
-            qml.PauliX,
-            qml.PauliY,
-            qml.PauliZ,
-            qml.S,
-            qml.T,
-            qml.RX,
-            qml.RY,
-            qml.RZ,
-            qml.Hadamard,
-            qml.Rot,
-            qml.CRot,
-            qml.Toffoli,
-            qml.SWAP,
-            qml.CSWAP,
-            qml.U1,
-            qml.U2,
-            qml.U3,
-            qml.CRX,
-            qml.CRY,
-            qml.CRZ,
+            qml.PauliX(wires=0),
+            qml.PauliY(wires=1),
+            qml.PauliZ(wires=2),
+            qml.S(wires=3),
+            qml.T(wires=0),
+            qml.RX(2.3, wires=1),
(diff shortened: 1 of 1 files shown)


tests/test_mixed_simulator_device.py
```diff
@@ -740,21 +740,11 @@ def circuit():
...
-                    )
+                    qml.apply(gate)
             return qml.expval(qml.PauliZ(0))
 
         qnode = qml.QNode(circuit, dev)
-        assert np.allclose(qnode(), 0.01160413)
-
-
-
+        assert np.allclose(qnode(), 0.0)
```
antalszava, line 747: How come the analytic value changed?

rmoyard: Because the parameters were given randomly before

tests/test_mixed_simulator_device.py
```diff
@@ -708,26 +708,26 @@ def test_four_qubit_random_circuit(self, shots):
...
-            qml.U3,
-            qml.CRX,
-            qml.CRY,
-            qml.CRZ,
+            qml.PauliX(wires=0),
+            qml.PauliY(wires=1),
+            qml.PauliZ(wires=2),
+            qml.S(wires=3),
+            qml.T(wires=0),
+            qml.RX(2.3, wires=1),
```
antalszava, line 716: How were the gate parameters and the wires chosen? Do they correspond to the previously used gate parameters/wires?

antalszava (commented):
Hi @rmoyard, looks good, just had two minor questions before approval.

tests/test_mixed_simulator_device.py
rmoyard: No I took what Maria used in the example for integration tests.

tests/test_mixed_simulator_device.py
antalszava: Right! We did "choose" parameters after all because we seeded, correct? If we test the same behaviour, then new parameters should be good :+1:

rmoyard: Yes and also because we cannot find the number of parameters that are necessary, so it is not possible to construct random circuit anymore.

antalszava (approved):
Looks good :rocket: Thanks! :slightly_smiling_face:

PR #89, opened 2022-01-26
[OpRefactor] Fix inverses

mariaschuld:
The operator refactor in PennyLane removed an obsolete legacy attribute `string_for_inverse`, which was set to `".inv"` everywhere and only used in one place in the core code.

However, the cirq device uses it in one place. Replacing it with `".inv"` should make the device compatible with the refactor, and have no downsides. While this is not urgent for the upcoming release, it would be easy to merge it in asap.

**Changes** 

Replace attribute call with fixed string.

Three minor improvements to pacify codefactor on unrelated issues, which are picked up in the latest version of the standard.

### DIFF
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -117,7 +117,7 @@ class CirqDevice(QubitDevice, abc.ABC):
             inverted_operation = CirqOperation(value.parametrization)
             inverted_operation.inv()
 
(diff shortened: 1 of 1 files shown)


mariaschuld:
@josh146, tagging you only for visibility

josh146 (approved):

josh146:
@antalszava @mariaschuld I think this can be merged in at any time!

antalszava (approved):
:tada:

PR #97, opened 2022-03-23
Update links to point to the PennyLaneAI GitHub organization

Jaybsoni:
Some links of PennyLane and the PennyLane plugin were still pointing to the XanaduAI organization instead of PennyLaneAI. These are now updated and the link for the blog points to the PennyLane blog.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -292,11 +292,11 @@ html_theme_options = {
     "table_header_border": "#19b37b",
     "download_button": "#19b37b",
     # gallery options
-    # "github_repo": "XanaduAI/PennyLane",
+    # "github_repo": "PennyLaneAI/PennyLane",
     # "gallery_dirs": "tutorials",
 }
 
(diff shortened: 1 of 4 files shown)


doc/conf.py
```diff
@@ -292,11 +292,11 @@ class TypeMock(type):
     "table_header_border": "#19b37b",
     "download_button": "#19b37b",
     # gallery options
-    # "github_repo": "XanaduAI/PennyLane",
+    # "github_repo": "PennyLaneAI/PennyLane",
```
rmoyard, line 295: Those are commented lines, should it be removed?

rmoyard (approved):
It looks good to me 👍 Just one comment on commented lines

doc/conf.py
Jaybsoni: Not sure, I know Antal had it changed aswell ..? Or is it a temporary comment for some reason?

PR #101, opened 2022-04-19
Cirq changed `initial_state` kwarg to be density matrix

Jaybsoni:
When calling the `simulator.simulate(circuit)` method for `cirq.DensityMatrixSimulator()` simulator (and all mixed state simulators I assume), the kwarg `initial_state=None` was previously passed as a state vector. Now it is required that this quantity be passed as a density matrix. 

The unit tests were updated to make use of this functionality. No other changes required

### DIFF
--- a/pennylane_cirq/simulator_device.py
+++ b/pennylane_cirq/simulator_device.py
@@ -118,7 +118,7 @@ class SimulatorDevice(CirqDevice):
         # get indices for which the state is changed to input state vector elements
         ravelled_indices = np.ravel_multi_index(unravelled_indices.T, [2] * self.num_wires)
 
-        state = np.zeros([2 ** self.num_wires], dtype=np.complex64)
+        state = np.zeros([2**self.num_wires], dtype=np.complex64)
         state[ravelled_indices] = state_vector
         state_vector = state.reshape([2] * self.num_wires)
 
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-pennylane>=0.17
+git+https://github.com/PennyLaneAI/pennylane@master#egg=pennylane
```
Jaybsoni, line 1:
```suggestion
pennylane>=0.17
```

pennylane_cirq/simulator_device.py
```diff
@@ -180,6 +187,11 @@ def analytic_probability(self, wires=None):
         probs = self._get_computational_basis_probs()
         return self.marginal_prob(probs, wires)
 
+    def _convert_to_density_matrix(self, state_vec):
+        """Convert ``state_vec`` into a density matrix."""
+        dim = 2**self.num_wires
+        return np.kron(state_vec, state_vec.conj()).reshape((dim, dim))
+
```
Jaybsoni, line 194: This is moved, but the `MixedStateSimulatorDevice` inherits from the base `SimulatorDevice` so both classes have access to this helper function.

rmoyard (approved):
Thank you for the fix of the tests, it looks good to me! 💯

antalszava:
[sc-17246]

PR #107, opened 2022-06-08
Define `returns_state` in `capabilities`

antalszava:
The Cirq devices that define the `state` method are capable of returning the state, yet they don't have the `returns_state=True` entry in the `capabilities` dictionary of the device (see [the explanation of `capabilities` here](https://pennylane.readthedocs.io/en/stable/development/plugins.html#device-capabilities)).


requirements-ci.txt
```diff
@@ -1,4 +1,4 @@
-pennylane
+git+git://github.com/PennyLaneAI/pennylane
```
rmoyard, line 1:
```suggestion
git+https://github.com/PennyLaneAI/pennylane.git
```

rmoyard (approved):
Thank you @antalszava, it looks good to me 💯

PR #114, opened 2022-09-19
support iswap, siswap, and their adjoints

timmysilv:
As the title says. These gates have native counterparts in cirq. I will work on making general adjoint support for plugins once all the native gates needing work are supported right out of the gate (hehehe)


albi3ro (approved):
Looks good 👍 

I wonder if we could define the `Adjoint(S)` and `Adjoint(T)` as instances of the `cirq.ZPowGate`? But that's something we can further look into later.

timmysilv:
I too think that's the way! I'm planning on using those PowGate classes to implement a generic solution for adjoints and powers after this PR

PR #115, opened 2022-09-21
support for ops raised to a power; remove old custom ops

timmysilv:
Introduce support for operations raised to a power using Cirq PowGates.

<details><summary>List of Cirq PowGates:</summary>

```bash
➜  ops pwd                                       
/Users/matthews/.pyenv/versions/3.9.13/envs/pl_plugins/lib/python3.9/site-packages/cirq_core-1.1.0.dev20220917023638-py3.9.egg/cirq/ops
➜  ops grep -ohrE "class \w+PowGate" .                    
class SwapPowGate
class ISwapPowGate
class CustomXPowGate
class XPowGate
class YPowGate
class ZPowGate
class HPowGate
class CZPowGate
class CXPowGate
class XXPowGate
class YYPowGate
class ZZPowGate
class CCZPowGate
class CCXPowGate
class PhasedISwapPowGate
class PhasedXPowGate
class WeightedZPowGate
```
</details>

As well, pennylane_cirq supports gates that exist in Cirq but not in PennyLane. This included CPhase and ISWAP, but those exist in PennyLane! I've removed the non-PL operation support for those and ensured the PL operation support.


pennylane_cirq/cirq_device.py
```diff
@@ -170,6 +182,17 @@ def __init__(self, wires, shots, qubits=None):
         "Projector": CirqOperation(lambda: cirq.ProductState.projector),
     }
 
+    def supports_operation(self, operation):
+        # pylint: disable=missing-function-docstring
+        op = operation
+        if isinstance(op, type) and issubclass(op, Operation):
+            op = op.name
```
AlbertMitjans, line 189: Careful! `name` is a property, thus it cannot be used as a class method:
```pycon
>>> qml.ops.Prod.name
<property at 0x10fc6e400>
```

pennylane_cirq/cirq_device.py
```diff
@@ -170,6 +182,17 @@ def __init__(self, wires, shots, qubits=None):
...
+        if isinstance(op, str):
+            op_with_power = op.split("**")
+            if len(op_with_power) == 2 and "Pow_" + op_with_power[0] in self._pow_operation_map:
+                return True
```
AlbertMitjans, line 193: To avoid creating more branches, we could just do sth like:
```suggestion
        if isinstance(op, str) and len(op.split("**")) == 2:
            operation = "Pow_" + op_with_power[0]
```
And then the check will be done in `super().supports_operation`

AlbertMitjans: Taking a deeper look at the code, it is weird that `self._complete_operation_map` is only used in `self._apply_operation` and not used in `self.operations` (which is the list used to see if the operation is supported or not).

Shouldn't we do:
```python
@property
def operations(self):
    return set(self._complete_operation_map.keys())
```

pennylane_cirq/cirq_device.py
AlbertMitjans: Will operation ever be a type (a `Pow` class)? If this happens we can't know if the operation is supported or not, given that since it is not an instance we don't know what the base is.

tests/test_cirq_device.py
```diff
@@ -156,6 +156,22 @@ def test_outer_init_of_qubits_with_wire_label_ints(self):
...
 
+    @pytest.mark.parametrize(
+        "op,expected",
+        [
+            (qml.PauliX, True),
+            ((qml.PauliX(0)**1.1).name, True),
+            ("PauliX**1.1", True),
+            (qml.IsingXX, False),
+            ("IsingXX", False)
+        ],
```
AlbertMitjans, line 167: Why are we adding a test for:
- `qml.PauliX`
- `qml.IsingXX`
These were already supported, right? They were not tested?

AlbertMitjans (commented):
Good job! Left some comments

pennylane_cirq/cirq_device.py
albi3ro: So I'm not quite sure when an type is passed.  `dev.stopping_condition` and `dev.check_validity` both use the string name.

pennylane_cirq/cirq_device.py
timmysilv: Your refactor makes me split the string twice which I don't love, but I'll think of a better way.

I didn't update to use `_complete_operation_map` because the `supports_operation` method is typically invoked with `obj.name`, so something like `qml.PauliX(0)**1.1` will ultimately be validated with `supports_operation("PauliX**1.1")` instead of `"Pow(PauliX)"` (which would match the `Adjoint(S)` pattern used above). Unless we changed the `qml.ops.Pow.name` property to something like `Pow(PauliX)`, we need to do something "hacky" on the plugin side to support powers. And I don't think we should change the name.

CHANGELOG.md
```diff
@@ -13,6 +16,10 @@
   the simulator not to ignore idle qubits. 
   [(#111)](https://github.com/PennyLaneAI/pennylane-cirq/pull/111) 
 
+* Removed the unnecessary `CPhase` and `ISWAP` operations from the
+  plugin's custom support because PennyLane supports them.
+  [(#115)](https://github.com/PennyLaneAI/pennylane-cirq/pull/115)
+
```
albi3ro, line 22: Should stick this under "breaking changes".

albi3ro (approved):
Nice solution! Thanks for this PR :)

pennylane_cirq/cirq_device.py
AlbertMitjans: I didn't mean to remove the `supports_operation` method overriding, I like this. But by using `_complete_operation_map` in `self.operations` you could change the operation name to `"Pow(PauliX)"` and then send it to `super().supports_operation`.

Something like:
```python
def supports_operation(self, operation):
    if isinstance(operation, str):
        op_with_power = operation.split("**")
        operation = "Pow_" + op_with_power[0] if len(op_with_power) == 2 else operation
    return super().supports_operation(operation)
"""

AlbertMitjans (approved):
Good job!

PR #123, opened 2022-11-28
Remove Python 3.7, add Python 3.10

albi3ro:
This PR adds Python 3.10 tests and removes Python 3.7.  Python 3.11 is not yet added as `qsimcirq` does not yet support Python 3.11.


timmysilv (approved):
nice! i'd be ok to wait for 3.11 if you don't want to change it twice, but this looks good

PR #130, opened 2023-02-11
remove inverse support; add adjoint support; state-prep op cleanup

timmysilv:
As the title says. `op.inv()` and `op.inverse` are removed from Operators in PennyLane 0.29, so this plugin needs updating accordingly. To summarize how I do it:
- All natively supported ops/obs are in a dict called `_base_op/obs_map`
- I dynamically copy the list, surrounding the names with `Adjoint(...)` (if the op exists and isn't already a natively supported adjoint op) and call the `_op/obs_map`. They're separate because the QSim devices did not support inverse ops so this will make it easy to preserve that assumption.
- The ops/obs list exposed to the pennylane `supports_op/obs` method now includes those `Adjoint(...)` ops/obs. No custom `device.capabilities["supports_inverse_operation"]` changes are needed because adjoint op names will be in the dict (rather than a hidden `.inv` property)
- The qsim(h) devices only expose the base ops (ie. without the Adjoint counterparts) to the pennylane `supports_obs/ops` thanks to the dict separation described above

The plugin also did lots of custom handling of StatePrep operators that will now be handled by the operators themselves, so I have removed that code. The custom handling had its own errors so tests were failing as they were, which is why I changed them in this PR. Both this and the inverse changes should happen in the same PR to update to the latest pennylane without breaking anything.

Note that because `StatePrep.state_vector` did not exist before PL 0.29, the version in setup.py will need to be bumped before this plugin is released.


timmysilv:
[sc-33452]

pennylane_cirq/cirq_device.py
```diff
@@ -171,7 +152,16 @@ def __init__(self, wires, shots, qubits=None):
         "Toffoli": CirqOperation(lambda: cirq.TOFFOLI),
     }
 
-    _observable_map = {
+    _operation_map = {
+        **_base_operation_map,
+        **{
+            f"Adjoint({key})": CirqOperation(value.parametrization, adjoint=True)
+            for key, value in _base_operation_map.items()
+            if value and key[:8] != "Adjoint("
```
AlbertMitjans, line 160: What does `if value` do?

AlbertMitjans: Do you know if:
```python
CirqOperation(CirqOperation(lambda: cirq.ISWAP).parametrization, adjoint=True)
```
is the same as:
```python
CirqOperation(lambda: cirq.ISWAP_INV)
```
?

We then could remove all the `Adjoint` operations from `_base_operation_map` and remove the string check.

tests/test_apply.py
```diff
@@ -91,51 +91,11 @@ def mimic_execution_for_apply(device):
...
-        [
-            np.array([0, 0]),
-            np.array([1, 0]),
-            np.array([0, 1]),
-            np.array([1, 1]),
-        ],
-    )
-    @pytest.mark.parametrize("device_wires", [3, 4, 5])
-    @pytest.mark.parametrize("op_wires", [[0, 1], [1, 0], [2, 0]])
-    def test_basis_state_on_wires_subset(self, state, device_wires, op_wires, tol):
```
AlbertMitjans, line 119: Why did you remove this test?

tests/test_apply.py
```diff
@@ -91,51 +91,11 @@ def mimic_execution_for_apply(device):
...
-        res = np.abs(dev.state) ** 2
-        # compute expected probabilities
-        expected = np.zeros([2 ** len(op_wires)])
-        expected[np.ravel_multi_index(state, [2] * len(op_wires))] = 1
-
-        expected = dev._expand_state(expected, op_wires)
-
-        assert np.allclose(res, expected, **tol)
-
-    def test_identity_basis_state(self, shots, tol):
```
AlbertMitjans, line 135: Same

tests/test_apply.py
AlbertMitjans: It would be nice to check that the behaviour hasn't changed

AlbertMitjans (approved):
Amazing job! Left some comments, but it is good to go 💯 🔥

tests/test_apply.py
timmysilv: It tests that BasisState parameters are expanded by `_expand_state` in this plugin, but it's not done in this plugin anymore! It's in `BasisState.state_vector()` which is in pennylane itself. This test would be double-coverage which is fine but imo unnecessary

tests/test_apply.py
timmysilv: This one is actually identical to `test_basis_state` above, but with a different state. So I parametrized that test to squish them into one and avoid copy-pasted tests 😋

tests/test_apply.py
AlbertMitjans: Did you at least test it locally? I wanted to make sure that the deleted implementation here was exactly the same (covered the exact same corner cases) than the implementation in `BasisState.state_vector()`

timmysilv: I did indeed run the tests before removing them, and I just triple-checked locally by ensuring the `_expand_state`, `dev._state` and `BasisState.state_prep()` all match each other on master. I could copy some test cases to [test_BasisState_state_vector](https://github.com/PennyLaneAI/pennylane/blob/023e001cd437e5283a4ffa1b2f73bda38ebf4bc9/tests/ops/qubit/test_state_prep.py#L167) as well if we're worried about coverage?

lillian542 (approved):
Looks good to me!

tests/test_apply.py
AlbertMitjans: I believe copying the tests over to PL is the safest choice 👍

timmysilv: moved them over in the linked PR below, lmk if we're good to go!

AlbertMitjans: Good to go! 🔥

PR #133, opened 2023-02-21
add PR link to changelog; require PL master

timmysilv:
fixing the changelog and requiring PL master. This is necessary because this release of pennylane_cirq uses `StatePrep.state_vector` which will be introduced in 0.29.


rmoyard (commented):
You can use "pennylane @ git+https://github.com/PennyLaneAI/pennylane.git" in setup.py instead and come back to it after the release and change to 0.29.0

rmoyard (approved):
Thanks, it looks good to me :+1:

PR #134, opened 2023-02-21
manual pre-release 0.29 updates

timmysilv:
just realized post-release was missed after 0.28, despite me manually making the changelog update in a previous PR! anyway - this is all that is missing for 0.29 to be release-ready now.

example of previous pre-release: #128 
changelog for this release added in #130 (first commit there)


Jaybsoni (approved):
Looks good!

PR #145, opened 2023-07-06
Remove do queue

lillian542:
The `do_queue` argument has been removed from PennyLane. Here we remove `do_queue=False` from the tests for the plugin.


frederikwilde (approved):
Looks good to me. I don't really understand why these `do_queue=False` statements were here. None of these operators are created in a queuing context anyways, no?

PR #150, opened 2023-09-21
remove expval ovverride; allow atol for cirq mixed state simulator test

timmysilv:
With numpy <1.25, the probs are identical (`[0.49999967, 0.49999967]`) before dotting with the eigenvalues so the result is exactly zero. With numpy >=1.25, they are not (`[0.49999964, 0.49999958]`). This slight change is very mysterious to me, but I have confirmed that it appears at [this line](https://github.com/PennyLaneAI/pennylane-cirq/blob/8447c1116355de8390789c2cb7794fa6deab50fd/pennylane_cirq/simulator_device.py#L109-L111). The inputted `self.circuit` is identical, and the outputted result is reliably off. Each value in the state is off by up to 3e-8, and I don't know why (they aren't totally random rounding errors because the difference between old and new states are numbers like `1.00000e-8` or `9.3400000e-9`).

My solution for now is to allow an atol of 5e-8.

EDIT: I've removed the expval override since it's not covered anymore. It was needed with the old numpy - should I pin the minimum, or keep the expval and put a no-cover comment on it?

EDIT 2: I decided to keep the override so we can still use old numpy without issues, and I added a test that forces use of the old one for codecov happiness

### DIFF
--- a/pennylane_cirq/simulator_device.py
+++ b/pennylane_cirq/simulator_device.py
@@ -244,16 +244,6 @@ class MixedStateSimulatorDevice(SimulatorDevice):
         )
         return capabilities
 
-    def expval(self, observable, shot_range=None, bin_size=None):
-        # The simulate_expectation_values from Cirq for mixed states involves
-        # a density matrix check, which does not always pass because the tolerance
-        # is too low. If the error is raised we use the PennyLane function for
-        # expectation value.
-        try:
-            return super().expval(observable, shot_range, bin_size)
-        except ValueError:
-            return qml.QubitDevice.expval(self, observable, shot_range, bin_size)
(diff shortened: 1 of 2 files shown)


mudit2812 (approved):
Looks good. I still think we should start thinking about unpinning Cirq

timmysilv:
@mudit2812 Christina is looking into that in another PR

codecov is failing because of a fallback mechanism that should probably be stated as no-cover, and perhaps will be hit less frequently with this numpy bump? I saw the issue locally in both scenarios so I'm not sure what's going on...

timmysilv:
updated the description with an EDIT, pls check it out

PR #151, opened 2023-10-02
Fix batch execute with pasqal device

albi3ro:
`cirq.Circuit` no longer has a device argument.

This wasn't caught as the reset method was not being tested for the pasqal device.  This PR also adds a test that will execute multiple circuits.

I also started converting `control_radius` to a float so we can start using `cirq.pasqal` in the automatic test matrix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,6 +83,7 @@ jobs:
           pl-device-test --device=cirq.simulator --tb=short --skip-ops --analytic=False --shots=20000 --cov=pennylane_cirq --cov-report=xml
           pl-device-test --device=cirq.mixedsimulator --tb=short --skip-ops --analytic=True --cov=pennylane_cirq --cov-report=xml
           pl-device-test --device=cirq.mixedsimulator --tb=short --skip-ops --analytic=False --shots=20000 --cov=pennylane_cirq --cov-report=xml
+          pl-device-test --device=cirq.pasqal --tb=short --skip-ops --analytic=False --shots=20000 --device-kwargs control_radius=2.
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,10 +12,14 @@
 
 ### Bug fixes 🐛
 
+* Fixes the pasqal device when more than one circuit is executed.
```
albi3ro, line 15:
```suggestion
* Fixes the pasqal device when more than one circuit is executed.
  [(#151)](https://github.com/PennyLaneAI/pennylane-cirq/pull/151)
```

.github/workflows/tests.yml
```diff
@@ -83,6 +83,7 @@ jobs:
           pl-device-test --device=cirq.simulator --tb=short --skip-ops --analytic=False --shots=20000 --cov=pennylane_cirq --cov-report=xml
           pl-device-test --device=cirq.mixedsimulator --tb=short --skip-ops --analytic=True --cov=pennylane_cirq --cov-report=xml
           pl-device-test --device=cirq.mixedsimulator --tb=short --skip-ops --analytic=False --shots=20000 --cov=pennylane_cirq --cov-report=xml
+          pl-device-test --device=cirq.pasqal --tb=short --skip-ops --analytic=False --shots=20000 --device-kwargs control_radius=2.
```
mudit2812, line 86: Why are the coverage flags not set?

mudit2812 (commented):
Just a question about coverage in the updated workflow

timmysilv (approved):
should we add a test for `reset` on that device now too?

.github/workflows/tests.yml
albi3ro:
```suggestion
          pl-device-test --device=cirq.pasqal --tb=short --skip-ops --analytic=False --shots=20000 --device-kwargs control_radius=2. --cov=pennylane_cirq --cov-report=xml
```

becuase I copy and pasted this from the plugin test matrix.  thanks for catching that.

albi3ro:
> should we add a test for `reset` on that device now too?

covered now by testing `batch_execute` with a batch.  Also by putting it in the device test suite.

albi3ro:
[sc-46688]

PR #158, opened 2023-11-07
stop using obs.return type in tests

timmysilv:
`qml._device.Device.sample` now works with `MeasurementProcess` objects, so I'm using those in tests instead to avoid the deprecation warning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Improvements 🛠
 
+* Sample tests have been updated to not use the deprecated `Observable.return_type` property.
+  [(#)]()
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Improvements 🛠
 
+* Sample tests have been updated to not use the deprecated `Observable.return_type` property.
+  [(#)]()
```
timmysilv, line 8:
```suggestion
  [(#158)](https://github.com/PennyLaneAI/pennylane-cirq/pull/158)
```

timmysilv:
#159 is better

PR #159, opened 2023-11-07
remove unnecessary `_obs_queue` property

albi3ro:
Tests in `test_sample.py` are setting an `_obs_queue` property and setting the return type of the observable.  

This private property might be a legacy holdover from several years ago. It does nothing at this current point in time.

So I'm simply removing them from the tests.


timmysilv (approved):
add to changelog and lgtm!

PR #186, opened 2024-04-22
BugFix: support for Identity on multiple wires

astralcai:
**Context**
`apply` fails with `Identity` on multiple wires because `cirq.I` only accepts a single wire.

**Change**
Replace `cirq.I` with `cirq.IdentityGate` which does allow specifying the number of wires

**Shortcut**
[sc-61681]

### DIFF
--- a/pennylane_cirq/cirq_device.py
+++ b/pennylane_cirq/cirq_device.py
@@ -171,7 +171,7 @@ class CirqDevice(QubitDevice, abc.ABC):
         "Hermitian": None,
         # TODO: Consider using qml.utils.decompose_hamiltonian() to support this observable.
         "Prod": None,
-        "Identity": CirqOperation(lambda: cirq.I),
+        "Identity": CirqOperation(lambda *args: cirq.IdentityGate(*args)),
         "Projector": CirqOperation(lambda: cirq.ProductState.projector),
     }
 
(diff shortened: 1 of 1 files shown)


pennylane_cirq/cirq_device.py
```diff
@@ -200,7 +200,12 @@ def to_paulistring(self, observable):
         cirq_op = self._observable_map[observable.name]
         if cirq_op is None:
             raise NotImplementedError(f"{observable.name} is not currently supported.")
-        cirq_op.parametrize(*observable.parameters)
+        parameters = (
+            [len(observable.wires)]
```
Qottmann, line 204: Why are the parameteres `[len(I.wires)]` in the case of an identity?

astralcai: That's just how `cirq.IdentityGate` is defined.

astralcai: For more context, in the cirq plugin, the gates must be parameterized first then applied on the different wires. Identity isn't technically a parameterized gate, but unlike other non parameterized gates which can be directly applied because they all act on a fixed number of wires, the Identity gate must be first initialized with the number of wires it expects.

Qottmann: Thanks!

Qottmann (approved):
Thanks @astralcai !

PR #211, opened 2025-03-17
Bump PL unpinning NP

JerryChen97:
PL is about to unpinning numpy.
`cirq` used to be super old and might contain some code incompatible. We would like to test with CI running.


JerryChen97:
Even though the CI still installed 1.26 instead of 2.1, all the tests passed without issue. All good.

PR #226, opened 2025-08-13
Use `set_shots` in tests

JerryChen97:
Besides, added a fixture `device_none_shots` to handle the test suites in `test_expval.py`
[sc-95432]

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,6 +111,7 @@ def device_analytic(request):
     device = request.param
 
     def _device(n):
-            return device(wires=n, shots=None)
+            return device(wires=n)
+
 
     return _device
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -96,17 +96,21 @@ def init_state(scope="session"):
 
 @pytest.fixture(params=analytic_devices + hw_devices)
 def device(request, shots):
-    """Fixture to initialize and return a PennyLane device"""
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -96,17 +96,21 @@ def _init_state(n):
...
             return device(wires=n, shots=shots)
 
-    elif device in hw_devices:
+    return _device
+
 
-        def _device(n):
-            return device(wires=n, shots=shots)
+@pytest.fixture(params=analytic_devices + hw_devices)
+def device_none_shots(request):
```
andrijapau, line 109: Maybe `analytic_device` is better?

tests/conftest.py
```diff
@@ -96,17 +96,21 @@ def _init_state(n):
...
 
-        def _device(n):
-            return device(wires=n, shots=shots)
+@pytest.fixture(params=analytic_devices + hw_devices)
+def device_analytic(request):
+    """Fixture to initialize and return a PennyLane device without shots"""
+    device = request.param
+
+    def _device(n):
+            return device(wires=n, shots=None)
```
andrijapau, line 114:
```suggestion
            return device(wires=n)

```
I guess we shouldn't even be using the shots= kwarg here at all right?

JerryChen97:
@andrijapau

